PR #99, opened 2022-02-08
Feature/arc 399

ian-emsens-shd:
https://meemoo.atlassian.net/browse/ARC-399


src/modules/shared/components/SearchBar/SearchBar.tsx
```diff
@@ -99,8 +100,9 @@ const SearchBar = <IsMulti extends boolean>({
 	 */
 
 	const rootCls = clsx(className, 'c-search-bar', {
-		['c-search-bar--lg']: large,
+		[`c-search-bar--${size}`]: size,
```
andrych, line 103: Is dit aangepast op de plaatsen die een large variant gebruiken?

src/pages/beheer/aanvragen/index.tsx
```diff
@@ -42,16 +43,41 @@ const CPRequestsPage: NextPage = () => {
...
+						variants={['rounded', 'light', 'bordered']}
+						onClick={(tabId) =>
+							setFilters({
+								...filters,
+								status: tabId.toString(),
+							})
+						}
+					/>
+
+					<SearchBar
```
bertyhell, line 59: The cursor is an arrow instead of a caret. This probably has to be fixed in the components library?
image

image

src/pages/beheer/aanvragen/index.tsx
```diff
@@ -42,16 +43,41 @@ const CPRequestsPage: NextPage = () => {
...
+						backspaceRemovesValue={false}
+						className="p-cp-requests__search"
+						instanceId="requests-seach-bar"
+						light={true}
```
bertyhell, line 63: FYI: you can omit the `={true}` if the value is true, since that is the default
```suggestion
						light
```

bertyhell (approved):

src/pages/beheer/aanvragen/index.tsx
andrych: Styling zit in de client dus kan hier gefixed worden.

andrych (approved):

PR #100, opened 2022-02-08
fix(arc-176): hide auth modal after login redirect

bertyhell:
Steps:
* currently when you go to the site as an unauthenticated user and click leeszaal 8 you get the auth modal
* login
* You are redirected to the leeszaal 8 detail page

current result:
You still see the auth modal after redirect
image

expected result:
You don't see the auth modal after the redirect
image


src/modules/auth/services/auth-service/auth.service.ts
```diff
@@ -23,7 +23,12 @@ class AuthService {
 	}
 
 	public redirectToLogin(query: StringifiableRecord) {
-		const returnToUrl = stringifyUrl({ url: config.public.origin, query });
+		// eslint-disable-next-line @typescript-eslint/no-unused-vars
+		const { showAuth, ...queryParams } = query;
```
andrych, line 27: Eventueel kunnen we in de `AuthModal` waar er geredirect wordt naar de login, eerst de `showAuth` query op false zetten?

ian-emsens-shd: https://github.com/viaacode/hetarchief-client/blob/master/src/pages/index.tsx#L140

```
			<AuthModal isOpen={showAuthModal && !user} onClose={onCloseAuthModal} />
```

bertyhell: good points

I've fixed both suggestions.

`{showAuthModal && !user}` hides the modal, but then the url still contains `http://localhost:3200/?showAuth=1` which is a bit confusing

andrych (approved):

PR #105, opened 2022-02-09
fix(fonts): fix baseline offset for sofia adobe fonts

bertyhell:
The font looks a bit thicker, but maybe we need to ask the designers to fix it on their side.

ezgif com-gif-maker-2


andrych (commented):
Ik zou hier nog even wachten tot team design antwoord geeft.

andrych (approved):

PR #116, opened 2022-02-11
fix(env-vars): fix env vars for tst

bertyhell:
- Fix missing origin env var (get it through publicRuntimeConfig instead of process.env)
- renamed NEXT_PUBLIC_ORIGIN to CLIENT_URL (more in line with PROXY_URL)
- Switch production check to BUILD_TOOLS === 'true' check (in case we ever want to enable build tools on production)
- Add better type checking for process.env, so we don't need to mirror env vars in a separate config.ts file

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -1,11 +1,11 @@
 # ENVIRONMENT
 NODE_ENV=local
 
-# SERVER
+# CLIENT PORT
 PORT=3200
 
-# CLIENT (Prefix with `NEXT_PUBLIC_`)
-NEXT_PUBLIC_ORIGIN=http://localhost:${PORT}
+# CLIENT URL
+CLIENT_URL=http://localhost:${PORT}
 
 # Next.js collects completely anonymous telemetry data about general usage.
 # Learn more here: https://nextjs.org/telemetry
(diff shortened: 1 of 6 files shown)


next.config.js
```diff
@@ -29,11 +29,13 @@ module.exports = withTM({
 	images: {
 		domains: ['assets.viaa.be'],
 	},
+	productionBrowserSourceMaps: process.env.DEBUG_TOOLS === 'true',
```
andrych, line 32: Dit gaat nu dus wel werken omdat die hieronder gedefinieerd staat?

.env.template
```diff
@@ -1,11 +1,11 @@
...
 NODE_ENV=local
 
-# SERVER
+# CLIENT PORT
 PORT=3200
 
-# CLIENT (Prefix with `NEXT_PUBLIC_`)
-NEXT_PUBLIC_ORIGIN=http://localhost:${PORT}
+# CLIENT URL
+CLIENT_URL=http://localhost:${PORT}
```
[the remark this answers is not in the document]
bertyhell: inderdaad, want PROXY_URL werkt ook zonder die prefix.

next.config.js
bertyhell: Will be provided by Tina: https://meemoo.atlassian.net/browse/ARC-484

andrych (approved):

PR #141, opened 2022-02-21
Feature/arc 186 profile page

andrych:
* Added profile page
* Refactored cp admin layout to re-usable sidebar layout
* Added other account pages and links in nav


src/modules/cp/const/index.ts
```diff
@@ -2,22 +2,22 @@ import { i18n } from 'next-i18next';
...
 	{
 		id: 'visitors',
-		label: i18n?.t('modules/cp/const/index___bezoekers'),
+		label: i18n?.t('modules/cp/const/index___bezoekers') ?? '',
 		href: '/beheer/bezoekers',
 	},
 	{
 		id: 'settings',
-		label: i18n?.t('modules/cp/const/index___instellingen'),
+		label: i18n?.t('modules/cp/const/index___instellingen') ?? '',
```
bertyhell, line 20: don't these automatically output the key if the translation value isn't found? So the `?? ''` case will never be used.

In avo we used i18next (not related to nextjs) to do the translations and that had an option to set values for keys that were not found:
```
		parseMissingKeyHandler: (key) => {
			if (key.includes('___')) {
				return `${upperFirst(lowerCase(key.split('___').pop()))} ***`;
			}
			return `${key} ***`;
		},
```

maybe nextjs has something similar?

src/pages/account/mijn-profiel/index.tsx
```diff
@@ -0,0 +1,65 @@
...
+							<dt>{t('pages/account/mijn-profiel/index___voornaam')}</dt>
+							<dd className="u-color-neutral">{user?.firstName}</dd>
+
+							<dt>{t('pages/account/mijn-profiel/index___familienaam')}</dt>
+							<dd className="u-color-neutral">{user?.lastName}</dd>
+
+							<dt>{t('pages/account/mijn-profiel/index___email')}</dt>
+							<dd className="u-text-ellipsis u-color-neutral" title={user?.email}>
+								{user?.email}
+							</dd>
```
bertyhell, line 46: some names can also be very long:
```suggestion
							<dt>{t('pages/account/mijn-profiel/index___voornaam')}</dt>
							<dd className="u-text-ellipsis u-color-neutral">{user?.firstName}</dd>

							<dt>{t('pages/account/mijn-profiel/index___familienaam')}</dt>
							<dd className="u-text-ellipsis u-color-neutral">{user?.lastName}</dd>

							<dt>{t('pages/account/mijn-profiel/index___email')}</dt>
							<dd className="u-text-ellipsis u-color-neutral" title={user?.email}>
								{user?.email}
							</dd>
```

bertyhell (changes requested):

src/modules/cp/const/index.ts
andrych: The problem is with the i18n instance from next-i18next, the type can be undefined and the `label` prop is expecting a string that's why I added the `?? ''` but I agree that it isn't the best solution. Didn't have the time to find a better solution.

src/pages/account/mijn-profiel/index.tsx
andrych: Fixed

ian-emsens-shd (approved):

PR #151, opened 2022-02-23
fix(arc-168): Split homepage into LoggedInHome and LoggedOutHome

bertyhell:
So we can wrap the loggedInHome component in a withAuth Higher OrderComponent.
This in turn ensures we keep all auth checking logic in one place.

Visually and functionally this PR should not change anything except for fixing a bug where the Terms of Service were not asked when someone logged into the site and was redirect to the homepage.


ian-emsens-shd (approved):

src/modules/home/components/LoggedInHome/__mocks__/VisitRequests.ts
```diff
@@ -12,6 +14,24 @@ export const heroLink = {
 
 export const heroImage = { src: '/images/hero.png', alt: 'hero image' };
 
+export interface HeroProps {
```
BavoVdg, line 17: Ik zou de interfaces in een `.types` file zetten ze sorteren voor `LoggedInHome` en `LoggedOutHome`. Idem voor de mocks.

PR #156, opened 2022-02-28
feat(arc-240): define media actions

BavoVdg:
Media acties gedefinieerd + extra _random_ acties om de `DynamicActionMenu` te testen.


bertyhell (approved):

bertyhell:
maybe we should also add some kind of tooltip for these buttons, since not it is a but of a guess for the user what they mean:
image

PR #165, opened 2022-03-07
Epic/collection interactions

ian-emsens-shd:
- Add remove and move actions in collection overview
- Add and integrate add-to-collection blade


bertyhell:
if the collections have no items and you try to bookmark an item, the app crashes:
image
image

ian-emsens-shd:
Collections will never be empty, the default folder will always exist as discussed here https://meemoo.atlassian.net/browse/ARC-212?focusedCommentId=29776

Regardless, I'll add a fallback

PR #166, opened 2022-03-07
Feature/arc 235

ian-emsens-shd:
- Add aggregations

Note: included in base branch under different commit


src/modules/media/hooks/get-media-objects.ts
```diff
@@ -2,21 +2,39 @@ import { useQuery } from 'react-query';
...
+			MediaService.getAll(rest, page, size),
+		]).then((responses) => {
+			const [results, noFormat] = responses;
+
+			return {
+				...results,
+				aggregations: noFormat.aggregations,
+			};
+		});
+	});
```
bertyhell, line 39:
```suggestion
	return useQuery(
		[QUERY_KEYS.getMediaObjects, { filters, page, size }],
		() => {
			const { format, ...rest } = filters || {};
			const mediaFormat: MediaTypes | undefined =
				format !== ReadingRoomMediaType.All ? format : undefined;

			// TODO: improve (?)
			// Run two queries to fetch aggregates across formats
			return Promise.all([
				MediaService.getAll({ ...rest, format: mediaFormat }, page, size),
				MediaService.getAll(rest, page, size),
			]).then((responses) => {
				const [results, noFormat] = responses;

				return {
					...results,
					aggregations: noFormat.aggregations,
				};
			});
		},
		{
			keepPreviousData: true,
		}
	);
```

bertyhell, line 39: Currently when you switch between tabs you get a flash of no results:
image

You can avoid this by telling react-query to keep the previous results until the new ones are loaded. We might need to add some kind of indicator that the results are loading though, since with this change it could be that the user doesn't get enough feedback that his click was accepted.

```suggestion
		},
		{
			keepPreviousData: true,
		}
	);
```

src/modules/media/hooks/get-media-objects.ts
```diff
@@ -2,21 +2,39 @@ import { useQuery } from 'react-query';
...
-	return useQuery([QUERY_KEYS.getMediaObjects, { filters, page, size }], () =>
-		MediaService.getAll(filters, page, size)
-	);
+	return useQuery([QUERY_KEYS.getMediaObjects, { filters, page, size }], () => {
+		const { format, ...rest } = filters || {};
+		const mediaFormat: MediaTypes | undefined =
+			format !== ReadingRoomMediaType.All ? format : undefined;
+
+		// TODO: improve (?)
+		// Run two queries to fetch aggregates across formats
```
bertyhell, line 27:
```suggestion
		// Run two queries:
		//     - One to fetch the aggregates across formats
		//     - And one to fetch te results for a specific tab (format)
```

bertyhell (approved):

BavoVdg:
ik krijg bij sommige zoekopdrachten wel aggregations, maar geen resultaten. Not sure if relevant here...
image
image

BavoVdg (approved):

PR #182, opened 2022-03-14
fix(arc-625): add flowplayer id and token

bertyhell:
before this PR: 
image

after this PR (hopefully)
image

The end vars that were added have already been added on TST, INT, QAS and PRD


andrych (approved):

BavoVdg (approved):
🤞

ian-emsens-shd (approved):

PR #185, opened 2022-03-15
feat(arc-508): show actual approved by column data in visitors table

bertyhell:
Requires these PR's for proxy and hasura (hasura is optional only needed for better seed data)
* https://github.com/viaacode/hetarchief-proxy/pull/91
* https://github.com/viaacode/hetarchief-hasura/pull/74

image

Also ensure times are shown in belgian time and not in utc time. Dates from the server are in format: 2021-12-30T23:00:00.000
But to parse them correctly we need to add a suffix: Z
eg: 2021-12-30T23:00:00.000Z

I've created a helper for this: formatAsBelgianDate

image


andrych (approved):

src/modules/shared/helpers/format-belgian-date.ts
```diff
@@ -0,0 +1,23 @@
+import { format } from 'date-fns';
+
+type DateFnsFormatOptions = {
```
andrych, line 3: Je kan deze ook bekomen via 
```typescript
type DateFnsFormatOptions = Parameters<typeof format>[2];
```

andrych (approved):

BavoVdg (approved):

PR #196, opened 2022-03-22
Fix/media endpoint

ian-emsens-shd:
- Expand metadata mapping with filters
- Align client with expected proxy filters
- Add some error-prevention on incorrect filters
- Adjust duration input to be able to match in Elastic


.npmrc
```diff
@@ -4,5 +4,8 @@ allows minor upgrades for that package, but after npm config set save-prefix='~'
 ~1.2.3 which only allows patch upgrades.
 save-prefix=""
 
+# required because the braft-editor still enforces react version 16 as a peer dependency
+legacy-peer-deps=true
+
```
bertyhell: This should no longer be needed since components version:
```
"@meemoo/react-components": "2.12.12",
```

Now the braft editor is correctly packaged by rollup so no requirements should be leaking out to the client npm install.

```suggestion
```

src/pages/leeszaal/[readingRoomSlug]/index.tsx
```diff
@@ -88,10 +96,31 @@ const ReadingRoomPage: NextPage = () => {
...
-		{
-			query: (query.search || []).join(' '),
-			format: (query.format as ReadingRoomMediaType) || READING_ROOM_QUERY_PARAM_INIT.format,
-		},
+		[
+			{
+				field: MediaSearchFilterField.QUERY,
+				operator: MediaSearchOperator.CONTAINS,
+				value: query.search !== null ? query.search?.toString() : '',
+			},
```
Comment (bertyhell) [inline src/pages/leeszaal/[readingRoomSlug]/index.tsx:104]: Only add the filter if the value is defined
```suggestion
			...(query.search?.toString()?.trim() ? [{
				field: MediaSearchFilterField.QUERY,
				operator: MediaSearchOperator.CONTAINS,
				value: query.search.toString(),
			}] : []),
```

src/pages/leeszaal/[readingRoomSlug]/index.tsx
```diff
@@ -88,10 +96,31 @@ const ReadingRoomPage: NextPage = () => {
...
+					const filters =
+						item.prop && item.op
+							? getMetadataSearchFilters(
+									item.prop as MetadataProp,
+									item.op as Operator
+							  )
+							: [];
+					return filters.map((filter, i) => ({ ...filter, value: values[i] }));
+				})
+				.reduce((prev, filters) => prev.concat(filters), []),
```
Comment (bertyhell) [inline src/pages/leeszaal/[readingRoomSlug]/index.tsx:122]: you might be able to just use a .flatMap() instead of .map().reduce()

src/pages/leeszaal/[readingRoomSlug]/index.tsx
Comment (ian-emsens-shd) [reply src/pages/leeszaal/[readingRoomSlug]/index.tsx:105]: This is handled in https://github.com/viaacode/hetarchief-client/blob/feature/ARC-663/src/modules/media/services/media/media.service.ts#L26-L47

src/modules/shared/types/index.ts
```diff
@@ -43,7 +43,7 @@ export enum Operator {
 	Equals = 'ac',
 	EqualsNot = 'ad',
 	LessThanOrEqual = 'ae', // shorter (duration) or until (date)
-	GreaterThan = 'af', // longer (duration) or after (date)
+	GreaterThanOrEqual = 'af', // longer (duration) or after (date)
 	Between = 'ag', // duration & date
```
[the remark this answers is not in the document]
ian-emsens-shd: Fixed in https://github.com/viaacode/hetarchief-client/commit/a05213db3c6d2c8bf1d24f9914baf7dddfc25dc8

.npmrc
ian-emsens-shd: Fixed in https://github.com/viaacode/hetarchief-client/commit/a05213db3c6d2c8bf1d24f9914baf7dddfc25dc8

src/pages/leeszaal/[readingRoomSlug]/index.tsx
Comment (ian-emsens-shd) [reply src/pages/leeszaal/[readingRoomSlug]/index.tsx:140]: Fixed in https://github.com/viaacode/hetarchief-client/commit/b2c853abdcbc58511d542fb9eaa669cb78f4ac78

PR #197, opened 2022-03-22
feat(arc-392): Show active readingRooms and search within 1 readingRoom

bertyhell:
Requires proxy PR: https://github.com/viaacode/hetarchief-proxy/pull/102

https://meemoo.atlassian.net/browse/ARC-392

image
image


src/pages/leeszaal/[readingRoomSlug]/index.tsx
```diff
@@ -57,6 +58,8 @@ import { VisitorLayout } from 'modules/visitors';
 
 const ReadingRoomPage: NextPage = () => {
 	const { t } = useTranslation();
+	const router = useRouter();
+	const { readingRoomSlug } = router.query;
```
Comment (ian-emsens-shd) [inline src/pages/leeszaal/[readingRoomSlug]/index.tsx:62]: Shouldn't this be `router.path`?

ian-emsens-shd (approved):

src/pages/leeszaal/[readingRoomSlug]/index.tsx
Comment (bertyhell) [reply src/pages/leeszaal/[readingRoomSlug]/index.tsx:73]: no router.query contains both queryParams and route params:
https://nextjs.org/docs/routing/dynamic-routes

image

PR #198, opened 2022-03-22
Feature/arc 663

ian-emsens-shd:
- Add Genre, Language, Medium and Creator filters with new CheckboxList component


bertyhell (approved):

src/modules/media/hooks/get-media-objects.ts
```diff
@@ -18,23 +18,28 @@ export function useGetMediaObjects(
...
 		[QUERY_KEYS.getMediaObjects, { filters, page, size, sort }],
 		() => {
-			// TODO: improve (?)
-			// Run two queries:
-			//     - One to fetch the results for a specific tab (format)
-			//     - and one to fetch the aggregates across formats
+			// TODO: improve ⚠️
+			// Run three queries:
+			//     - One to fetch the results for a specific tab (results),
+			//     - one to fetch the aggregates without any criteria to populate filters (noFilters)
```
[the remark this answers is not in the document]
ian-emsens-shd: Filter logic is OR, not AND, filter values are not allowed to vanish when refining results.

src/modules/reading-room/const/index.tsx
```diff
@@ -26,7 +26,7 @@ export const READING_ROOM_ITEM_COUNT = 100;
 export const READING_ROOM_QUERY_PARAM_INIT = {
 	format: ReadingRoomMediaType.All,
 	orderProp: ReadingRoomSort.Relevance,
-	page: 1,
+	page: 0,
```
bertyhell, line 29: All page params in the app start counting from 1.
The backend also uses this rule.

Should this one be different for some reason?

bertyhell (approved):

src/modules/reading-room/const/index.tsx
ian-emsens-shd: Pagination was non-functional because it's 0-indexed, reverted and adjusted there

PR #255, opened 2022-04-19
feat(arc-490): implement redirect for register user button in auth modal

bertyhell:
Add register flow:
image
image
image
image
image

### DIFF
--- a/src/modules/auth/components/AuthModal/AuthModal.tsx
+++ b/src/modules/auth/components/AuthModal/AuthModal.tsx
@@ -25,6 +25,12 @@ const AuthModal: FC<AuthModalProps> = (props) => {
 		await AuthService.redirectToLoginHetArchief(queryParams, router);
 	};
 
+	const onRegisterHetArchief = async () => {
+		// eslint-disable-next-line @typescript-eslint/no-unused-vars
+		const { showAuth, ...queryParams } = query;
+		await AuthService.redirectToRegisterHetArchief(queryParams, router);
+	};
+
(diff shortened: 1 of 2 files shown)


src/modules/auth/components/AuthModal/AuthModal.tsx
```diff
@@ -25,6 +25,12 @@ const AuthModal: FC<AuthModalProps> = (props) => {
 		await AuthService.redirectToLoginHetArchief(queryParams, router);
 	};
 
+	const onRegisterHetArchief = async () => {
+		// eslint-disable-next-line @typescript-eslint/no-unused-vars
+		const { showAuth, ...queryParams } = query;
+		await AuthService.redirectToRegisterHetArchief(queryParams, router);
+	};
+
 	const onLoginMeemoo = async () => {
```
ian-emsens-shd, line 34: Consistentie in naamgeving, `...HetArchief` vs `...Meemoo`?

ian-emsens-shd (approved):

src/modules/auth/components/AuthModal/AuthModal.tsx
bertyhell: there are 2 identity providers: 
* hetarchief 
* meemoo

so there are also 2 login methods:
* onLoginHetArchief
* onLoginMeemoo

New visitors can only register in the hetarchief identity provider (idp), so there is only one register handler:
* onRegisterHetArchief

I think that is consistent. Do you agree?

PR #256, opened 2022-04-19
feat(arc-512): load navigation items from backend

bertyhell:
Requires hasura seed PR: https://github.com/viaacode/hetarchief-hasura/pull/114
Requires proxy PR: https://github.com/viaacode/hetarchief-proxy/pull/149

notice: the jira item does not match with the implementation. This PR implements fetching navigation items from the database, so they can be used by the frontend to show in the header navigation bar and in the center of the footer

image
image


src/modules/navigation/components/Navigation/Navigation.consts.tsx
```diff
@@ -54,142 +58,121 @@ const renderLink = (
...
-		children: [
-			{
-				node: renderLink('Alle leeszalen', '/', {
-					className: dropdownCls(['u-display-none', 'u-display-block:md']),
+	accessibleReadingRooms: ReadingRoomInfo[],
+	navigationItems: NavigationInfo[]
+): NavigationItem[] => {
+	return [
+		{
+			node: renderLink('Bezoekersruimtes', '', {
```
ian-emsens-shd, line 66: i18n

src/modules/navigation/services/navigation-service/navigation.types.ts
```diff
@@ -1,6 +1,19 @@
-export interface Navigation {
-	description: string;
+export enum NavigationPlacement {
+	HeaderLeft = 'header_left',
+	FooterCenter = 'footer_center',
+}
+
+export interface NavigationInfo {
```
ian-emsens-shd, line 6: Personally not a fan of the `Info`-suffix, doesn't make the name more descriptive imo

ian-emsens-shd (approved):

src/modules/navigation/services/navigation-service/navigation.types.ts
bertyhell: agreed, this is mostly done because there is already a Navigation component, and IDE's would then try to import the wrong one.
Also in the proxy there was a naming conflicts between what we load from the database "Navigation" and what we send to the client "NavigationInfo". So i kept the name consistent between proxy and client.

PR #260, opened 2022-04-20
Bugfix/arc 391

ian-emsens-shd:
- avoid nav pushing beyond page size
- adjust spacing
- ensure image is portrait, regardless of height

### DIFF
--- a/src/modules/navigation/components/Navigation/Navigation.module.scss
+++ b/src/modules/navigation/components/Navigation/Navigation.module.scss
@@ -199,10 +199,10 @@ $c-navigation-dropdown-item-spacer: 1.2rem;
 	.c-navigation__list-overlay {
 		width: 100vw;
 		position: absolute;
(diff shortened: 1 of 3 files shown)


bertyhell:
In the future, try to include a screenshot of styling fixes, otherwise these are very hard to PR-review
also include some words in the branch name, it makes it easier to find the branch without having to lookup the issue id
eg: 
```bugfix/ARC-391-navigation-styling-fixes```
instead of 
```bugfix/ARC-391```

bertyhell:
In the future, try to include a screenshot of styling fixes, otherwise these are very hard to PR-review
also include some words in the branch name, it makes it easier to find the branch without having to lookup the issue id
eg: 
```bugfix/ARC-391-navigation-styling-fixes```
instead of 
```bugfix/ARC-391```

bertyhell (approved):

PR #266, opened 2022-04-21
Bugfix/arc 630

ian-emsens-shd:
Resolve some scroll issues, minor i18n, have withAuth pull from state

### DIFF
--- a/src/modules/auth/wrappers/with-auth/with-auth.tsx
+++ b/src/modules/auth/wrappers/with-auth/with-auth.tsx
@@ -1,8 +1,10 @@
 import { useRouter } from 'next/router';
 import { stringify } from 'query-string';
-import { ComponentType, useCallback, useEffect, useState } from 'react';
+import { ComponentType, useCallback, useEffect } from 'react';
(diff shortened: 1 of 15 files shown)


src/modules/auth/wrappers/with-auth/with-auth.tsx
```diff
@@ -60,10 +59,11 @@ export const withAuth = (
...
 
-		if (!isLoggedIn) {
-			return <Loading fullscreen />;
-		}
-
-		return <WrappedComponent {...props} />;
+		return (
+			<>
+				{!isLoggedIn && <Loading fullscreen />}
+				<WrappedComponent {...props} />
```
bertyhell, line 65: doesn't this risk loading the spinner and the component at the same time?

bertyhell (approved):

src/modules/auth/wrappers/with-auth/with-auth.tsx
ian-emsens-shd: Intentional; this prevents white-flash-of-content when the logged in status re-checks.

PR #270, opened 2022-04-22
feat(arc-448): Show 404 page if space is not accessible or doesn't exist

bertyhell:
Requires backend PR: https://github.com/viaacode/hetarchief-proxy/pull/161

## Search page of a space

Should show a no access when the space exist, but the user does not have access (active visit request)
image

Should show 404 page when space does not exist:
image

Should show search page when space exists and the user has access:
image

## Detail page

Should show "not found" error if the slug for a space is not found but the object detail schema_identifier is found:
image

should show "no access" error if you don't have access to the space, even when going to the detail page
image

should show 'not found" error if the schema_identifier of the object is not found
image

should show object detail if user has access to space, space exists and schema_identifier of the object is found:
image

## Error boundaries

Also added error boundaries so the app doesn't crash completely:
image

also in beheer, so the sidebar stays up:
image


src/pages/[slug]/[ie]/index.tsx
```diff
@@ -636,6 +644,29 @@ const ObjectDetailPage: NextPage = () => {
...
+	const renderPageContent = () => {
+		if (mediaInfoIsLoading || visitRequestIsLoading) {
+			return <Loading fullscreen />;
+		}
+		if (isErrorNotFound) {
+			return <ErrorNotFound />;
+		}
+		if (isErrorSpaceNoAccess) {
+			return <ErrorSpaceNoAccess visitorSpaceSlug={router.query.slug as string} />;
+		}
```
Comment (ian-emsens-shd) [inline src/pages/[slug]/[ie]/index.tsx:659]: Waarom zijn dit geen redirects naar `404`- & `403`-pagina's?

src/pages/[slug]/[ie]/index.tsx
```diff
@@ -636,6 +644,29 @@ const ObjectDetailPage: NextPage = () => {
...
+		return <div className="p-object-detail">{renderObjectDetail()}</div>;
+	};
+
+	return (
+		<VisitorLayout>
+			<Head>
+				<title>{createPageTitle('Object detail')}</title>
+				<meta name="description" content="Object detail omschrijving" />
```
Comment (ian-emsens-shd) [inline src/pages/[slug]/[ie]/index.tsx:667]: i18n

ian-emsens-shd (changes requested):

src/pages/[slug]/[ie]/index.tsx
Comment (bertyhell) [reply src/pages/[slug]/[ie]/index.tsx:659]: Dan verliest de gebruiker de url waar hij wilde naar toe gaan.
bv: hij wil een url in typen en maakt een typo.
Dan krijg je:
bezoek.hetarchief.be/huis-van-alin => bezoek.hetarchief.be/404

En dan moet de gebruiker die volledige opnieuw intypen

ian-emsens-shd (approved):

PR #280, opened 2022-04-25
feat(arc-303): limit kiosk functionalities + create logout route

bertyhell:
requires proxy PR: https://github.com/viaacode/hetarchief-proxy/pull/166

* limit functionalities for kiosk permissions
* add logout route

Remove export button on folder overview: EXPORT_OBJECT
image

Remove bookmark and export button on object detail: EXPORT_OBJECT, MANAGE_FOLDERS
image

Remove bookmark button on search page: MANAGE_FOLDERS
image

Remove profile and notifications + shield /account routes + redirect to homepage
image


src/modules/auth/store/user/user.select.ts
```diff
@@ -9,7 +8,3 @@ export const selectIsLoggedIn = (state: AppState): boolean => !!state.user.user;
 export const selectHasAcceptedTosAt = (state: AppState): string | null => {
 	return state.user.user?.acceptedTosAt || null;
 };
-
-export const selectHasPermission = (state: AppState, permission: Permission): boolean => {
-	return !!state.user.user?.permissions.includes(permission);
-};
```
ian-emsens-shd, line 15: Why remove this selector? You could still wrap it in a hook
```
export const selectHasAllPermissions = (state: AppState, permission: Permission): boolean => {
	return state.user.user && intersection(state.user.user.permissions.permissions, permissions).length === permissions.length;
};
```
```
export const useHasAllPermission = (...permissions: Permission[]): boolean => {
	return useSelector(selectHasAllPermissions);
};
```

src/pages/[slug]/index.tsx
```diff
@@ -417,6 +416,26 @@ const ReadingRoomPage: NextPage = () => {
 		);
 	};
 
+	const getCardButtons = (item: MediaCardProps): ReactNode => {
```
Comment (ian-emsens-shd) [inline src/pages/[slug]/index.tsx:419]:
```suggestion
	const renderCardButtons = (item: MediaCardProps): ReactNode => {
```

src/pages/account/mijn-historiek/index.tsx
```diff
@@ -27,6 +31,18 @@ import { VisitorLayout } from 'modules/visitors';
...
 	const { t } = useTranslation();
 	const [filters, setFilters] = useQueryParams(ACCOUNT_HISTORY_QUERY_PARAM_CONFIG);
+	const canManageAccount: boolean | null = useHasAllPermission(Permission.MANAGE_ACCOUNT);
+	const router = useRouter();
+
+	useEffect(() => {
+		if (!canManageAccount) {
+			toastService.notify({
+				title: t('Geen toegang'),
+				description: t('Je hebt geen rechten om deze pagina te bekijken'),
```
ian-emsens-shd, line 41: i18n

ian-emsens-shd (changes requested):

src/modules/auth/store/user/user.select.ts
bertyhell: Since i created hooks specifically to get these permissions, we don't need the selector anymore. Better to just have 1 way to fetch permissions.
https://github.com/viaacode/hetarchief-client/pull/280/files#diff-71792b997b9f74a7d73fe70e8a0d8ce483adbf9a5bee1a09e900cfd2cbe910a0R11-R17

bertyhell:
consolidated into https://github.com/viaacode/hetarchief-client/pull/286

PR #283, opened 2022-04-25
fix(navigation): bezoekersruimtes active state

BavoVdg:
- Bezoekersruimtes active state

### DIFF
--- a/src/modules/navigation/components/Navigation/Navigation.consts.tsx
+++ b/src/modules/navigation/components/Navigation/Navigation.consts.tsx
@@ -9,7 +9,7 @@ import styles from '@navigation/components/Navigation/Navigation.module.scss';
 import { NavigationInfo } from '@navigation/services/navigation-service/navigation.types';
 import { VisitorSpaceInfo } from '@reading-room/types';
 import { Icon, IconName } from '@shared/components';
-import { ROUTES } from '@shared/const';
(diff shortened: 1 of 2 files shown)


ian-emsens-shd (approved):

src/modules/navigation/components/Navigation/Navigation.consts.tsx
```diff
@@ -90,7 +90,14 @@ export const getNavigationItemsLeft = (
...
 		),
 		id: 'visitor-spaces',
-		active: currentPath === ROUTES.home,
+		active:
+			currentPath === ROUTES.home ||
+			Object.keys(ROUTE_PREFIXES).every(
+				(key) =>
+					!currentPath.startsWith(
+						`/${(ROUTE_PREFIXES as { [key: string]: string })[key]}`
+					)
```
bertyhell, line 99: this works for now, but will not work once we add content pages since those can have any url.
We'll have to make this an async check, so that we can query the backend if this space exists, if it does, then this should be the active navigation item.

Let's merge as is, then fix it once we add content pages.

bertyhell (approved):

src/modules/navigation/components/Navigation/Navigation.consts.tsx
ian-emsens-shd: https://meemoo.atlassian.net/browse/ARC-790

PR #285, opened 2022-04-25
fix(arc-716): various fixes

BavoVdg:
- Metadata top border
- Show keywords if similar items do not exist
- Mobile spacing around content on object detail


bertyhell:
Add some whitespace below the keywords, so when the feedback button position is fixed, it doesn't overlap the keywords:
image

bertyhell (approved):

ian-emsens-shd (approved):

PR #293, opened 2022-04-26
Bugfix/arc 624

ian-emsens-shd:
- force TOS-gate

https://meemoo.atlassian.net/browse/ARC-624


BavoVdg (approved):

src/modules/auth/wrappers/with-auth/with-auth.tsx
```diff
@@ -61,7 +55,7 @@ export const withAuth = (
 
 		return (
 			<>
-				{!isLoggedIn && <Loading fullscreen />}
+				{!tosAccepted && <Loading fullscreen />}
 				<WrappedComponent {...props} />
 			</>
 		);
```
bertyhell, line 61: preferably you never render the wrappedComponent when the toc is not accepted yet.
Otherwise the wrapped component starts initializing and throwing errors since it does not have a valid session yet:

http://localhost:3200/vrt
image

```
if (!tosAccepted) {
	return <Loading fullscreen />
}
return <WrappedComponent {...props} />
```

PR #295, opened 2022-04-26
fix: improve type security and improve fallbacks

ian-emsens-shd:
Should address fatal errors on QAS for Christophe's user mentioned in https://studiohyperdrive.slack.com/archives/C0279QSL9CZ/p1650964466333869

- improve type security
- improve fallbacks

### DIFF
--- a/src/modules/home/components/LoggedInHome/LoggedInHome.tsx
+++ b/src/modules/home/components/LoggedInHome/LoggedInHome.tsx
@@ -13,7 +13,7 @@ import ReadingRoomCardsWithSearch from '@home/components/ReadingRoomCardsWithSea
 import { VISITOR_SPACE_SLUG_QUERY_KEY } from '@home/const';
 import { useCreateVisitRequest } from '@home/hooks/create-visit-request';
 import { useGetReadingRoom } from '@reading-room/hooks/get-reading-room';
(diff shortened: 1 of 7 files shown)


src/modules/home/components/LoggedInHome/LoggedInHome.tsx
```diff
@@ -194,13 +194,14 @@ const LoggedInHome: FC = () => {
...
-		info: visit.spaceInfo,
+	const mapVisitToRoom = (visit: Visit): VisitorSpaceCardProps['room'] => ({
+		image: visit.spaceImage || null,
+		logo: visit.spaceLogo || '[ERR] No logo found',
+		color: visit.spaceColor || null,
+		name: visit.spaceName || '[ERR] No name found',
+		info: visit.spaceInfo || '[ERR] No info found',
 		slug: visit.spaceSlug,
+		id: visit.spaceId,
 	});
```
bertyhell, line 205: This sets values for the VisitorSpaceCardProps['room'] object that cannot be distinguished from actual urls.
If they are undefined, set them to undefined
and make sure the code that uses them can deal with undefined values. You can default to these string values inside the component, but not inside intermediate objects that are passed around.

This feels a bit like string 'undefined' values, which can get very messy.

bertyhell (changes requested):

PR #299, opened 2022-04-26
feat(arc-161): Add cookie policy page

bertyhell:
This doesn't seem to load correctly on localhost
but maybe it will on QAS just like the cookie popup did

```
consentcdn.cookiebot.com/consentconfig/e17bca33-78a0-484e-a204-e05274a65598/localhost/configuration.js:1          Failed to load resource: the server responded with a status of 404 ()
```

image


src/pages/cookiebeleid/index.tsx
```diff
@@ -0,0 +1,19 @@
+import { NextPage } from 'next';
+import Script from 'next/script';
+import React from 'react';
+
+import styles from './cookie-policy.module.scss';
+
+const CookiePolicy: NextPage = () => {
+	return (
+		<div className={styles['cookie-policy__wrapper']}>
```
ian-emsens-shd, line 9: Use a (probably Sticky) layout and `p-`-prefix for page-level classes

ian-emsens-shd (changes requested):

src/pages/cookiebeleid/index.tsx
bertyhell: The cookie policy page should look something like this?
https://onderwijs-qas.hetarchief.be/cookiebeleid

How would a sticky layout work for that?

PR #310, opened 2022-04-28
fix(arc-708): Correctly render new lines in description and abstract

bertyhell:
You can test this by manually updating an object description in the database:
http://localhost:3200/vrt/49b1bf8894004fd49aeaba36cfc5a958d5c32a4566244999a862e80b498a2c7c7bee152896204294938534fc7f3c6793
image

image
image

### DIFF
--- a/src/modules/media/utils/render-with-new-lines.tsx
+++ b/src/modules/media/utils/render-with-new-lines.tsx
@@ -1,0 +1,24 @@
+import DOMPurify from 'dompurify';
+import { ReactNode } from 'react';
+
+import { RICH_TEXT_SANITIZATION } from '@shared/const';
+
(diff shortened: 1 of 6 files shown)


src/modules/media/utils/render-with-new-lines.tsx
```diff
@@ -0,0 +1,24 @@
...
+					DOMPurify.sanitize(
+						// Replace new lines and literal new lines (description and abstract are encoded differently)
+						text.replaceAll('\\n', '<br/><br/>').replaceAll('\n', '<br/><br/>'),
+						RICH_TEXT_SANITIZATION
+					)
+				),
+			}}
+		/>
+	);
+}
```
ian-emsens-shd, line 24: Isn't this essentially a component? Not a util?

PR #311, opened 2022-04-28
fix(arc-555): Scroll to top of page on every pagination page change

bertyhell:
image
image


src/modules/shared/utils/scroll-to-top.ts
```diff
@@ -0,0 +1,3 @@
+export function scrollToTop(): void {
+	window.scrollTo(0, 0);
```
ian-emsens-shd, line 2:
```suggestion
	window.scrollTo({ top: 0, left: 0, behavior: 'smooth' });
```

ian-emsens-shd (changes requested):

PR #320, opened 2022-04-29
Feature/arc 266

ian-emsens-shd:
- Expand media state
- Add collections to media state
- Rework features that need collections to fetch from state
- Adjust icon type based on business logic

https://meemoo.atlassian.net/browse/ARC-266


bertyhell:
Works for search page, but not for detail page. Should detail page also be part of this PR?
image

image

src/modules/account/types/collections.ts
```diff
@@ -1,4 +1,6 @@
-import { MediaTypes } from '@shared/types';
+import { ApiResponseWrapper, MediaTypes } from '@shared/types';
+
+export type GetCollections = ApiResponseWrapper<Collection>;
```
bertyhell, line 3: this sounds like it is a type of a function, not an object
maybe GetCollectionsResponse is better?

src/modules/reading-room/components/AddToCollectionBlade/AddToCollectionBlade.tsx
```diff
@@ -281,7 +284,7 @@ const AddToCollectionBlade: FC<AddToCollectionBladeProps> = (props) => {
 
 								<li className={styles['c-add-to-collection-blade__list-button']}>
 									<CreateCollectionButton
-										afterSubmit={() => collections.refetch()}
+										afterSubmit={() => getCollections.refetch()}
```
bertyhell, line 287: can be simplified, or does typescript compain when you do this?
```suggestion
										afterSubmit={getCollections.refetch}
```

bertyhell (approved):

PR #337, opened 2022-05-03
Bugfix/arc 743

ian-emsens-shd:
- Fix resetting advanced form
- Avoid being able to set empty advanced filters

https://meemoo.atlassian.net/browse/ARC-743


bertyhell:
Advanced filter stays empty after going back to it:
image

bertyhell (changes requested):
see other remark

PR #339, opened 2022-05-03
fix(media-card): set image placeholder height to 16:9 for nicer loading

bertyhell:
Currently when loading the media on the search page, before the images load, a square white space is reserved.
After loading the images that space collapses to a 16 by 9 size.

image

image

By setting the image size to be 16:9 by default, we get a nicer loading behavior.

image

For the images that are not 16:9, the nextimage is set to be fill, so it will just be cropped to 16:9.

### DIFF
--- a/src/modules/shared/components/MediaCard/MediaCard.module.scss
+++ b/src/modules/shared/components/MediaCard/MediaCard.module.scss
@@ -29,7 +29,7 @@ $c-media-card-no-video-intended-spacing: 0.9rem;
 
 				> img {
 					position: static !important;
-					height: auto !important;
+					height: calc(238px / 16 * 9) !important;
 				}
 			}
 		}


src/modules/shared/components/MediaCard/MediaCard.module.scss
```diff
@@ -29,7 +29,7 @@ $c-media-card-no-video-intended-spacing: 0.9rem;
 
 				> img {
 					position: static !important;
-					height: auto !important;
+					height: calc(238px / 16 * 9) !important;
```
ian-emsens-shd, line 32: `238px` in een variabele steken en toepassen waar nodig + responsiviteit is niet in orde.
image
image
image
Ik zou deze PR gewoon close'n tbh, de moeite niet waard.

ian-emsens-shd (changes requested):

bertyhell:
too much work to get it working on all column widths

PR #346, opened 2022-05-04
fix(arc-787): logout when api returns 401 unauthorized

bartnaessens:
logout when api returns 401 unauthorized.

Related proxy PR: https://github.com/viaacode/hetarchief-proxy/pull/181

### DIFF
--- a/src/modules/shared/services/api-service/api.service.ts
+++ b/src/modules/shared/services/api-service/api.service.ts
@@ -2,6 +2,8 @@ import ky from 'ky-universal';
 import { KyInstance } from 'ky/distribution/types/ky';
 import getConfig from 'next/config';
 
+import { AuthService } from '@auth/services/auth-service';
+
 const { publicRuntimeConfig } = getConfig();
 
 export abstract class ApiService {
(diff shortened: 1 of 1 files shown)


src/modules/shared/services/api-service/api.service.ts
```diff
@@ -15,6 +17,16 @@ export abstract class ApiService {
 					'content-type': 'application/json',
 				},
 				credentials: 'include', // TODO change to same-origin once working on server
+				hooks: {
+					afterResponse: [
+						(_request, _options, response) => {
+							if (response.status === 401) {
+								// user is unauthorized and should login again
+								AuthService.logout();
```
bertyhell, line 25: In this case it might be nice if after the user logs back in, they are redirected back to the page they were trying to visit:
```
	public static logout(shouldRedirectToOriginalPage: boolean): void {
		let returnToUrl = publicRuntimeConfig.CLIENT_URL;
		if (shouldRedirectToOriginalPage) {
			const path = window.location.href.substring(publicRuntimeConfig.CLIENT_URL.length);
			returnToUrl = stringifyUrl({
				url: publicRuntimeConfig.CLIENT_URL,
				query: {
					redirectTo: path,
					showAuth: 1
				}
			});
		}

		window.location.href = stringifyUrl({
			url: `${publicRuntimeConfig.PROXY_URL}/auth/global-logout`,
			query: {
				returnToUrl,
			},
		});
	}
```

bertyhell (changes requested):

ian-emsens-shd (approved):

BavoVdg (approved):

src/modules/shared/services/api-service/api.service.ts
bartnaessens: fixed

PR #358, opened 2022-05-06
feat(arc-513): improve search performance + reduce empty state between

bertyhell:
https://meemoo.atlassian.net/browse/ARC-513?focusedCommentId=31294

* Move "get aggregations" to separate hook, so it is only loaded once and not on every search
* keep previous results + filter type, so the results from the "all" tab do not show audio objects in the results of the video tab when switching to video
* Don't make 2 queries to the elasticsearch when filters is empty, since the queries are identical
* Set size to 0 when loading aggregations, since we do not care about the results, only the aggregations
* Use MediaSearchFilter[] instead of MediaSearchFilters, since it makes it clearer that it's an array and not an object

before:
https://user-images.githubusercontent.com/1710840/167099895-f9ec41ce-459c-49bd-b3d6-eb0dba5c153d.mp4

after:
https://user-images.githubusercontent.com/1710840/167099937-2da0f8e8-c8ac-487a-845f-aa66942e1bf4.mp4


src/pages/[slug]/index.tsx
```diff
@@ -445,6 +456,13 @@ const ReadingRoomPage: NextPage = () => {
 		<>
 			<MediaCardList
 				items={media?.items
+					.filter((mediaObject: MediaInfo) => {
+						if (query.format && query.format !== 'all') {
+							return mediaObject.dcterms_format === query.format;
+						} else {
+							return true;
+						}
+					})
```
Comment (ian-emsens-shd) [inline src/pages/[slug]/index.tsx:465]: Avoid filtering front-end as it messes with pagination-UX

ian-emsens-shd (changes requested):

ian-emsens-shd (approved):

src/pages/[slug]/index.tsx
Comment (bertyhell) [reply src/pages/[slug]/index.tsx]: yes, this isn't for the real results, but for the intermediate state between switching tabs

when you are on the "all" tab, you load some results (video + film + audio)
when you click on the video tab, you want to do a new query, without losing your tab count data
but if i don't filter the results, you'll see a flash of the old audio items in the video tab, which looks very wrong.

This avoids that flash and also keeps the tab count giving it a smoother and seemingly faster result

PR #410, opened 2022-05-31
fix(arc-898): show loading message until visit status has loaded

bertyhell:
https://user-images.githubusercontent.com/1710840/171130871-316d6328-e167-4241-947b-97931b3ddeec.mp4


BavoVdg (changes requested):
Minor: naming

PR #455, opened 2022-06-08
fix(arc-981): reset page when filters/sorting changes on search page

bertyhell:
https://meemoo.atlassian.net/jira/software/c/projects/ARC/issues/ARC-981

cannot be tested locally because we do not have enough search results.

The problem:
Go to search on QAS en goto page 4
image

Scroll back to the top and add a filter:
image

Scroll back down:
image

Current result: the page is still on page 4

Desired result: the page is reset to page 1


BavoVdg (approved):
Getest door met QAS db te verbinden

PR #457, opened 2022-06-08
fix(arc-545): replace custom class with variant for content info button

bertyhell:


bertyhell:
forgot to push one PR remark fix

PR #489, opened 2022-06-15
fix(arc-1019): reduce tag selector height

bertyhell:
https://meemoo.atlassian.net/browse/ARC-1019

before:
image

after:
image
image

### DIFF
--- a/src/styles/pages/admin/_rewrites.scss
+++ b/src/styles/pages/admin/_rewrites.scss
@@ -209,10 +209,10 @@ button[class*="c-button"].c-checkbox-dropdown-modal__trigger {
 	}
 
 	/**
-	* Cross-browser styles for explicit focus via
-	* keyboard-based (eg Tab) navigation or the
-	* .focus-visible utility class.
-	*/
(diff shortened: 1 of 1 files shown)


src/styles/pages/admin/_rewrites.scss
```diff
@@ -515,6 +515,17 @@ button[class*="c-button"].c-checkbox-dropdown-modal__trigger {
 // dropdowns
 
 .c-filter-table__toolbar {
+	.c-dropdown__trigger {
+		.c-checkbox-dropdown-modal__trigger {
+			padding: 0.54rem $spacer-md;
+		}
+
+		[class^="TagList-module_c-tag-list"] {
```
ian-emsens-shd, line 523: Will this selector work after a PRD build's class obfuscation?

ian-emsens-shd (approved):

src/styles/pages/admin/_rewrites.scss
bertyhell: yes, the class always starts the same way:
image

PR #492, opened 2022-06-16
fix(arc-256): ensure content page title is below mobile hamburger menu

bertyhell:
https://meemoo.atlassian.net/browse/ARC-256

before:
image

after:
image


src/styles/layouts/_content-block-preview.scss
```diff
@@ -36,6 +34,10 @@
 	}
 }
 
+.c-content-page-preview .c-content-block > div {
+	z-index: 1; // Below mobile hamburger menu
```
ian-emsens-shd, line 38: Use https://github.com/viaacode/hetarchief-client/blob/master/src/styles/abstracts/_variables.scss#L108

ian-emsens-shd (changes requested):

PR #496, opened 2022-06-17
fix(arc-854): link visitor space back link to visitor space with search

bertyhell:
https://meemoo.atlassian.net/browse/ARC-854

image

You can test this by adding this code in the renderMetaData inside the pages/[slug]/[ie]/index file
```
{/*TODO remove test link*/}
<Link
	  passHref
	  href="/vrt/09f17b37445c4ce59f645c2d5db9dbf8dbee79eba623459caa8c6496108641a0900618cb6ceb4e9b8ad907e47b980ee3"
>
	<a>Link to another item</a>
</Link>
```

- Then you go to vrt search page
- search for "vragen"
- klik on the only item
- click on the fake link you just added
- click on the back button in the black header bar

current result:
you return to /vret without filters

expected result:
you return to the /vrt with vragen filter intact


ian-emsens-shd (changes requested):
⚠️ regression
`?focus=${router.query.ie}` is not covered by these changes; If I recall correctly this is the mechanism to return to the previous page and highlight/scroll-to the previous result.

bertyhell:
> ⚠️ regression
> `?focus=${router.query.ie}` is not covered by these changes; If I recall correctly this is the mechanism to return to the previous page and highlight/scroll-to the previous result.

Still works:
image

PR #507, opened 2022-08-05
fix(arc-1084): add waveform for related & similar

ian-emsens-shd:
https://meemoo.atlassian.net/browse/ARC-913


bartnaessens (approved):

bertyhell:
can't you use the same asset that we already have?
public/images/waveform.svg

Or does the color have to be different? Maybe there is a way to change the fill color using css?

image

public/images/waveform--white.svg
```diff
@@ -0,0 +1,158 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN" "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd">
+<svg version="1.1" id="Layer_1" xmlns="http://www.w3.org/2000/svg" x="0px" y="0px"
```
bertyhell, line 3: a green version of this waveform already exists, shouldn't we show the same one?

bertyhell (commented):
see global remarks

public/images/waveform--white.svg
ian-emsens-shd: image

Reworking the existing waveform to use currentColor produces too many potential side-effects

bertyhell (approved):

PR #538, opened 2022-08-24
fix(arc-1025): add and remove tab index on blade close button

bertyhell:
https://meemoo.atlassian.net/browse/ARC-1025

Requires components PR: https://github.com/viaacode/react-components/pull/66

This PR adds a hack to attempt to hide the outline on the close button of the notification center blade on ios after opening the profile dropdown.

This hack has not been tested since i couldn't get the vpn to work with browserstack local. 
@ian-emsens-shd if you manage to get it working while testing this PR, let me know the settings?

image


ian-emsens-shd (commented):
Adjusted the hack; does not influence tab functionality in the notification center.

Note: WCAG changes in PR should include tabindex'es for the notifications

PR #539, opened 2022-08-24
Feature/arc 951

ian-emsens-shd:
https://meemoo.atlassian.net/browse/ARC-951

### DIFF
--- a/src/modules/navigation/components/Navigation/NavigationList/NavigationList.tsx
+++ b/src/modules/navigation/components/Navigation/NavigationList/NavigationList.tsx
@@ -5,7 +5,6 @@ import { useSelector } from 'react-redux';
 
 import { Icon, IconLightNames, Overlay } from '@shared/components';
 import { selectHistory } from '@shared/store/history';
-import { scrollTo } from '@shared/utils/scroll-to-top';
 
 import styles from '../Navigation.module.scss';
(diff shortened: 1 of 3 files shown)


src/modules/navigation/components/Navigation/NavigationList/NavigationList.tsx
```diff
@@ -33,7 +34,12 @@ const NavigationList: FC<NavigationListProps> = ({ currentPath = '', items, onOp
 
 	const renderTrigger = (item: NavigationItem, iconName: IconLightNames) => {
 		return (
-			<div className={clsx(styles['c-navigation__link--wrapper'], 'u-cursor-pointer')}>
+			<div
+				className={clsx(styles['c-navigation__link--wrapper'], 'u-cursor-pointer')}
+				onKeyDown={(e) => onKey(e, keysSpacebar, () => e.preventDefault())}
```
bertyhell, line 39: to select an item in the menu i can press enter, but the space key does not work.
Shouldn't pressing space also activate the button?

ian-emsens-shd: Can't have both; preventing default scrolling behaviour also prevents default event bubble-up behaviour.

PR #544, opened 2022-08-30
Bugfix/arc 918

ian-emsens-shd:
https://meemoo.atlassian.net/browse/ARC-918


ian-emsens-shd:
Tested on INT; fatal error avoided on detail pages

bertyhell (approved):

PR #549, opened 2022-09-08
fix(avo-1084): remove left and right margin from card title wrapper

bertyhell:
https://meemoo.atlassian.net/browse/ARC-1084

before:
image

after:
image


ian-emsens-shd (approved):
Pushed a commit that avoids the below regression:
Screenshot 2022-09-09 at 14 50 04

PR #550, opened 2022-09-08
feat(arc-964): replace i18n with implementation that supports html

bertyhell:
https://meemoo.atlassian.net/browse/ARC-964

This clears the way for ARC-964 by allowing translations to contain html and rendering those translations as html where supported.
Locations where no html is supported use the `tText(): string` function instead of the `t(): string | ReactNode` function
If a transaltion does not contain any html, it is rendered as text

Locations that do not support html:
* labels of buttons
* aria labels
* pagina titels
* meta tag descriptions
* placeholders of input fields

Both relative and absolute links are supported. Although the full app will be reloaded when you click them. We might look into optimizing that using a Next Link in the future.

absolute url:
image

relative url:
image

Current admin translations screen:
image


ian-emsens-shd (changes requested):
Why are we reinventing the wheel instead of using existing solutions?

https://stackoverflow.com/a/52108122
https://stackoverflow.com/a/72177021

bertyhell:
> Why are we reinventing the wheel instead of using existing solutions?

It seems like i18next only supports basic html nodes like: strong, em, b. And links have to be added as <1></1> which then get mapped to the <a href></a> in the code.
But meemoo wants to set their own link, so we need actual html.

bertyhell:
merged into https://github.com/viaacode/hetarchief-client/pull/557

PR #552, opened 2022-09-09
feat(avo-964): add translations v2 admin screen from admin-core

bertyhell:
https://meemoo.atlassian.net/browse/ARC-964

also disabled stylelint rule: no-descending-specificity
Since they recommend disabling the rule when you use a lot of nesting
> We recommend turning this rule off if you use a lot of nesting.

https://stylelint.io/user-guide/rules/list/no-descending-specificity/

### DIFF
--- a/src/pages/admin/vertalingen-v2/index.tsx
+++ b/src/pages/admin/vertalingen-v2/index.tsx
@@ -1,0 +1,99 @@
+import { TranslationsOverviewV2 } from '@meemoo/react-admin';
+import { Button } from '@meemoo/react-components';
+import clsx from 'clsx';
+import Head from 'next/head';
+import React, { FC, ReactNode } from 'react';
+
+import { Permission } from '@account/const';
+import { AdminLayout } from '@admin/layouts';
+import { withAdminCoreConfig } from '@admin/wrappers/with-admin-core-config';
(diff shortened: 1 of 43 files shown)


src/pages/admin/vertalingen-v2/TranslationsOverviewV2.module.scss
```diff
@@ -0,0 +1,89 @@
...
+				font-size: 20px;
+				margin-top: 4px;
+			}
+		}
+	}
+}
+
+.c-translations-overview__blade:global(.c-blade--active) {
+	width: 700px;
+	max-width: 100%;
```
ian-emsens-shd, line 60: Blades are not allowed to fill the screen

src/pages/admin/vertalingen-v2/index.tsx
```diff
@@ -0,0 +1,100 @@
...
+				<div
+					className={clsx(
+						'u-px-32 u-py-24',
+						styles['c-translations-overview__blade-footer']
+					)}
+				>
+					<Button
+						variants={['block', 'black']}
+						onClick={onSave}
+						label={tText('Bewaar wijzigingen')}
```
ian-emsens-shd, line 45: Extract translations

ian-emsens-shd (changes requested):

PR #558, opened 2022-09-13
feat(arc-1095): add related items indicator

ian-emsens-shd:
https://meemoo.atlassian.net/browse/ARC-1069

Requires https://github.com/viaacode/hetarchief-proxy/pull/246


bertyhell:
works correctly:
image

src/modules/media/services/media/media.service.ts
```diff
@@ -115,6 +115,24 @@ export class MediaService {
...
 
+	public static async countRelated(
+		meemooIdentifiers: string[] = []
+	): Promise<Record<string, number>> {
+		return await ApiService.getApi()
+			.get(
+				stringifyUrl(
+					{
+						url: `${MEDIA_SERVICE_BASE_URL}/related/count`,
+						query: { meemooIdentifiers },
```
bertyhell, line 126: side lore:
not sure if you took this into account, but concatinating ids can sometimes give issues since the max url length is 2048 chars
lucky for you, the meemoo id is only 10 chars long
so 39 * 10 => 310 chars

But if you would build this endpoint with schema_identifiers which are 96 chars long you would run into issues:
96 * 39 => 3744 chars

src/modules/shared/types/media.ts
```diff
@@ -48,7 +48,8 @@ export interface MediaSearchFilter {
 
 export interface MediaInfo {
 	meemoo_fragment_id: string; // Unique id per object
-	schema_identifier: string; // PID (this is not unique per objects)
+	schema_identifier: string;
+	meemoo_identifier: string; // PID (this is not unique per objects)
```
bertyhell, line 52: I think meemoo_fragment_id no longer exists
So the code with comments becomes:
```typescript
	schema_identifier: string; // Unique id per object
	meemoo_identifier: string; // PID (this is not unique per objects)
```

I can see one more location it is used in the client: src/modules/media/services/media/media.service.ts
But i think this can also be switched to the schema_identifier:
```typescript
		return {
			items: parsed?.hits?.hits.map((item) => ({
				...item._source,
				schema_identifier: item._id,
			})),
			total: parsed?.hits?.total?.value,
			size: size,
			page: page,
			pages: Math.ceil((parsed?.hits?.total?.value || 0) / size),
			aggregations: parsed.aggregations,
		};
```

src/modules/shared/components/MediaCard/MediaCard.types.ts
```diff
@@ -17,6 +17,7 @@ export interface MediaCardProps {
 	title?: string | ReactNode;
 	type: MediaTypes;
 	view?: MediaCardViewMode;
+	hasChildren?: boolean;
```
bertyhell, line 20: maybe rename this to hasRelated
since we don't know if they are children until meemoo adds the parent child indicator. And hasChildren might be confusing when we implement the parent-child feature in the future.

bertyhell (approved):

src/modules/media/services/media/media.service.ts
ian-emsens-shd: wanted to keep it `GET` but yeah, long term we should look at an equivalent `POST`

PR #561, opened 2022-09-19
fix(arc-964): avoid module css in Html component

bertyhell:
since nextjs tries to serverside render pages, it loads the Html component
which in turn loads it's module scss,
which will try to add a style node to the document head, but document does not exist server side,
so it crashes and the page instead loads a 404 page

I'm not sure why this is not an issue with other module scss components

You can see the error by checking out develop and surfing to:
http://localhost:3200/account/mijn-mappen/favorieten

```
error - (middleware)\node_modules\next\dist\build\webpack\loaders\next-style-loader\runtime\injectStylesIntoStyleTag.js (80:0) @ <unknown>
ReferenceError: document is not defined
```

image

### DIFF
--- a/src/modules/shared/components/Html/Html.tsx
+++ b/src/modules/shared/components/Html/Html.tsx
@@ -25,7 +25,6 @@ const Html: FunctionComponent<HtmlProps> = ({
 				__html: DOMPurify.sanitize(content, sanitizePreset) as string,
 			}}
 			// Avoid module scss file, since next serverside rendering can't add a style block to the document.
-			style={{ overflowWrap: 'break-word' }}
 			className={clsx('c-html-wrapper', className)}
 		/>
 	);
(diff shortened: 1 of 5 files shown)


src/modules/shared/components/Html/Html.tsx
```diff
@@ -26,7 +24,9 @@ const Html: FunctionComponent<HtmlProps> = ({
 			dangerouslySetInnerHTML={{
 				__html: DOMPurify.sanitize(content, sanitizePreset) as string,
 			}}
-			className={clsx(styles['c-html-wrapper'], className)}
+			// Avoid module scss file, since next serverside rendering can't add a style block to the document.
+			style={{ overflowWrap: 'break-word' }}
```
ian-emsens-shd, line 28: Instead of defining an inline style; move this to the global .scss file

ian-emsens-shd (changes requested):

src/modules/shared/components/Html/Html.tsx
bertyhell: fixed, moved to: src/styles/components/_html.scss

ian-emsens-shd (approved):

PR #564, opened 2022-09-19
fix(arc-1099): add custom scrollbar for related objects on object detail

bertyhell:
https://meemoo.atlassian.net/browse/ARC-1099

### DIFF
--- a/src/styles/pages/_object-detail.scss
+++ b/src/styles/pages/_object-detail.scss
@@ -182,6 +182,10 @@ $metadata-collapsed-width: 36vw;
 
 	&__related {
 		max-width: 100vw;
+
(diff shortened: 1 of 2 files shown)


src/styles/pages/_object-detail.scss
```diff
@@ -182,6 +182,10 @@ $metadata-collapsed-width: 36vw;
 
 	&__related {
 		max-width: 100vw;
+
+		.ScrollbarsCustom-TrackY > .ScrollbarsCustom-Thumb {
```
ian-emsens-shd, line 186: You're already targetting the scrollbar here; avoid inline styles in the component

ian-emsens-shd (changes requested):

ian-emsens-shd (approved):

PR #565, opened 2022-09-19
fix(arc-634): increase line-height of media cards on the result page

bertyhell:
https://meemoo.atlassian.net/browse/ARC-634

ezgif com-gif-maker (4)


src/modules/shared/components/MediaCard/MediaCard.module.scss
```diff
@@ -46,6 +46,11 @@ $c-media-card-no-video-intended-spacing: 0.9rem;
 		}
 	}
 
+	:global(.c-card__subtitle-wrapper),
+	:global(.c-card__children-wrapper) {
+		line-height: 2.2rem;
+	}
```
ian-emsens-shd, line 52: This only changes the line-height by `1px`.
`font-size: 1.4rem; line-height: 1.5;` => `line-height: 21px`

This creates inconsistencies with other implementations of `Card` and deviates from the defined line-height's (`$line-height-4xl`) so I'd like to avoid it. I seem to recall we checked these minor inconsistencies during initial component development and were told it's fine to homogenise them based on the existing variables.

bertyhell: could be, i wasn't part of that discussion.
Can you add your remarks to the task and tag ineke?
https://meemoo.atlassian.net/browse/ARC-634

PR #571, opened 2022-09-21
fix(arc-904): switch user group select dropdown with checkboxes

bertyhell:
https://meemoo.atlassian.net/browse/ARC-904

requres admin-core PR: https://github.com/viaacode/react-admin-core-module/pull/43

before:
image
image
image
image

after:
image
image
image
image


src/styles/pages/admin/_content-pages.scss
```diff
@@ -1,6 +1,8 @@
 @use "src/styles/abstracts" as *;
 
 .c-admin__content {
+	padding-bottom: 10rem;
```
ian-emsens-shd, line 4: Use a previously defined `$spacer`
```
$spacer-1xl: 4rem;
$spacer-2xl: 8rem;
$spacer-3xl: 12rem;
```

src/styles/pages/admin/_content-pages.scss
```diff
@@ -47,4 +49,27 @@
...
+	.c-user-group-select {
+		.c-checkbox-group {
+			border: 1px solid grey;
+			border-radius: $input-border-radius;
+			background-color: $white;
+			padding: 1px;
+
+			.c-checkbox {
+				display: block;
+				padding: 1.5rem;
```
ian-emsens-shd, line 62: Removing the `1px` padding on the parent and upping the child's padding to `$spacer-sm` should have about the same effect right?

ian-emsens-shd (changes requested):

src/styles/pages/admin/_content-pages.scss
bertyhell: I've changed it to a $spacer-sm.
The 1px padding is needed to avoid the hover effect to cut into the parent border. like this:
image

ian-emsens-shd (approved):

PR #574, opened 2022-09-22
fix(arc-791): add margin right to icon in request visit cards

bertyhell:
https://meemoo.atlassian.net/browse/ARC-791

before:
image
image

after:
image
image


ian-emsens-shd (commented):
See pushed commit

PR #587, opened 2022-10-26
feat(VisitorSpaceSettings): fix deleting of image when editing image …

WardVerc1:
…and description


ian-emsens-shd (approved):
Not sure why but it works without regression

PR #592, opened 2022-10-27
fix(avo-542): fix server side rendering for object detail

bertyhell:
https://meemoo.atlassian.net/browse/ARC-542

requires proxy PR: https://github.com/viaacode/hetarchief-proxy/pull/259

example url: http://localhost:3200/vlaams-parlement/d7d7841f8ad946f8bfa5116b30b9041ccec8fc5521d2404cab08b86c2d41466e231c382d924942de8d586b2f713e85f5

Server side rendered html now correctly contains the SEO tags like title, description, og tags
image
image

Switches SEO from client side to server side rendering for the object detail page.
This is a proof of concept. If you guys approve with the approach, ill apply it for all pages.

### DIFF
--- a/src/modules/shared/helpers/render-og-tags.tsx
+++ b/src/modules/shared/helpers/render-og-tags.tsx
@@ -1,12 +1,15 @@
+import getConfig from 'next/config';
 import Head from 'next/head';
 import { ReactNode } from 'react';
 
 import { createPageTitle } from '@shared/utils';
 
+const { publicRuntimeConfig } = getConfig();
+
 export function renderOgTags(
 	title: string | undefined,
 	description: string,
-	clientUrl: string
+	url: string
 ): ReactNode {
 	const resolvedTitle = createPageTitle(title);
 	return (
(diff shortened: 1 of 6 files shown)


src/modules/shared/helpers/render-og-tags.tsx
```diff
@@ -1,25 +1,28 @@
...
+			<meta property="og:url" content={url} />
 			<meta property="og:title" content={resolvedTitle} />
 			<meta property="og:description" content={description} />
-			<meta property="og:image" content={`${clientUrl}/images/og.jpg`} />
+			<meta property="og:image" content={`${publicRuntimeConfig.CLIENT_URL}/images/og.jpg`} />
 			<meta property="twitter:card" content="summary_large_image" />
-			<meta property="twitter:domain" content={window.location.origin} />
+			<meta property="twitter:domain" content={publicRuntimeConfig.CLIENT_URL} />
 			<meta property="twitter:title" content={resolvedTitle} />
 			<meta property="twitter:description" content={description} />
```
ian-emsens-shd, line 27: Basically all of these fields will have to be changed per-page; to avoid parameter bloat I suggest switching to something like the below and providing fallback values.

```
function renderOgTags(config?: {
  // shared
  description?: string;
  
  // og-specific
  og?: {
    type?: string;
    ....
  },
  
  // twitter-specific
  twitter?: { ... }
}): ReactNode { ... }
```

ian-emsens-shd (approved):

src/modules/shared/helpers/render-og-tags.tsx
bertyhell: > Basically all of these fields will have to be changed per-page

Could be, well implemented it when they ask for it.

PR #595, opened 2022-10-28
feat(arc-542): render seo tags for all pages using server side rendering

bertyhell:
https://meemoo.atlassian.net/browse/ARC-542

you can view the source of a page using chrome by typing this in the url bar:
view-source:http://localhost:3200/

view-source:http://localhost:3200/vlaams-parlement/d7d7841f8ad946f8bfa5116b30b9041ccec8fc5521d2404cab08b86c2d41466e231c382d924942de8d586b2f713e85f5

Fixes server side rendering SEO for all routes

image

image

image

image

image

image


bertyhell:
@ian-emsens-shd ik merge deze al, maar als je die nog eens kan reviewen woensdag zou dat handig zijn

PR #600, opened 2022-12-07
feat(admin-core): switch to react-query v4 to match admin-core-ui

bertyhell:
Reduces code duplication by 52 lines
still todo: fix navigations/items fetching, currently returns empty array


src/modules/navigation/services/navigation-service/navigation.service.ts
```diff
@@ -2,16 +2,15 @@ import { ApiService } from '@shared/services/api-service';
...
+export class NavigationService {
+	private static baseUrl = 'admin/navigations/items';
 
-	public async getAll(): Promise<Record<NavigationPlacement, NavigationInfo[]>> {
+	public static async getAll(): Promise<Record<NavigationPlacement, NavigationInfo[]>> {
 		const response: Record<NavigationPlacement, NavigationInfo[]> = await ApiService.getApi()
 			.get(this.baseUrl)
 			.json();
 
+		console.log('getting navigation items', response);
```
WardVerc1, line 13: Remove console.log?

PR #601, opened 2023-01-02
feat(ARC-964): fix hyperlinks in translations

WardVerc1:
image

### DIFF
--- a/src/modules/media/components/ObjectPlaceholder/ObjectPlaceholder.types.ts
+++ b/src/modules/media/components/ObjectPlaceholder/ObjectPlaceholder.types.ts
@@ -1,9 +1,11 @@
+import { ReactNode } from 'react';
+
 import { DefaultComponentProps } from '@shared/types';
 
 export interface ObjectPlaceholderProps extends DefaultComponentProps {
-	description?: string;
+	description?: string | ReactNode;
 	reasonTitle?: string;
-	reasonDescription?: string;
+	reasonDescription?: string | ReactNode;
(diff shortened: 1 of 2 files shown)


src/modules/media/components/ObjectPlaceholder/ObjectPlaceholder.types.ts
```diff
@@ -1,9 +1,11 @@
+import { ReactNode } from 'react';
+
 import { DefaultComponentProps } from '@shared/types';
 
 export interface ObjectPlaceholderProps extends DefaultComponentProps {
-	description?: string;
+	description?: string | ReactNode;
```
ian-emsens-shd, line 6: `ReactNode` includes `string` already
```suggestion
	description?: ReactNode;
```

ian-emsens-shd (approved):

src/modules/media/components/ObjectPlaceholder/ObjectPlaceholder.types.ts
```diff
@@ -1,9 +1,11 @@
+import { ReactNode } from 'react';
+
 import { DefaultComponentProps } from '@shared/types';
 
 export interface ObjectPlaceholderProps extends DefaultComponentProps {
-	description?: string;
+	description?: ReactNode;
 	reasonTitle?: string;
-	reasonDescription?: string;
```
bertyhell, line 6: we should allow both string and ReactNode since the component can render both

bertyhell (approved):

PR #610, opened 2023-01-18
feature/ARC-1224: add filtering options for creation- and published date

WardVerc1:
image


src/modules/visitor-space/components/YearInput/YearInput.tsx
```diff
@@ -0,0 +1,27 @@
...
+			<b>{props.label}</b>
+		</div>
+		<TextInput
+			style={{ textAlign: 'right' }}
+			iconStart={<Icon name="calendar" />}
+			placeholder="jjjj"
+			type="number"
+			{...props}
+		/>
+	</div>
```
bertyhell, line 24: image

src/modules/visitor-space/components/CreatedFilterForm/CreatedFilterForm.tsx
```diff
@@ -74,84 +84,160 @@ const CreatedFilterForm: FC<CreatedFilterFormProps> = ({ children, className, di
...
+					id={labelKeys.created}
+					label={tHtml(
+						'modules/visitor-space/components/created-filter-form/created-filter-form___waarde'
+					)}
+				>
+					<SelectDateOrYear
+						yearsSelected={yearsSelected}
+						setYearsSelected={setYearsSelected}
+						showPluralLabel={showRange}
+					/>
```
bertyhell, line 186: if we enter a year: 2023
and apply
then reopen the filter. It shows "1 Jan. 2023" with date selected instead of year.

Maybe we should default to the year radio box if the selected date starts with "1 Jan." when we open de filter?

image

src/modules/visitor-space/components/SelectDateOrYear/SelectDateOrYear.module.scss
```diff
@@ -0,0 +1,11 @@
+$component: "radiobuttons-container";
+
+.#{$component} {
+	display: flex;
+	flex-direction: column;
+	margin: 30px 0;
+
+	&__radiobutton {
+		margin: 10px 0;
```
bertyhell, line 9: use rem instead of px

bertyhell (changes requested):

src/modules/visitor-space/components/CreatedFilterForm/CreatedFilterForm.tsx
```diff
@@ -74,84 +84,160 @@ const CreatedFilterForm: FC<CreatedFilterFormProps> = ({ children, className, di
...
-							}
-						/>
-					</FormControl>
-				</div>
+							) : (
+								<DateInput
+									label={getSelectValue(operators, form.operator)?.label}
+									disabled={disabled}
+									id={labelKeys.created}
+									onChange={(date) => {
```
silkederudder-shd, line 228: Maybe move this to a function above the render e.g. `onChangeDateInputChange`. This would make the render fn easier to read. Same for the other onChange functions.

src/modules/visitor-space/components/DateInput/DateInput.tsx
```diff
@@ -14,8 +14,15 @@ import styles from './DateInput.module.scss';
...
 
-const DateInput: FC<DatepickerProps> = (props) => (
+interface DateInputProps extends DatepickerProps {
+	label?: string;
+}
+
+const DateInput: FC<DateInputProps> = (props) => (
 	<div className={styles['c-date-input']}>
+		<div className={styles['c-date-input__label']}>
+			<b>{props.label}</b>
```
silkederudder-shd, line 24: You could also use p-tag and add font-weight styling ;)

silkederudder-shd (changes requested):

src/modules/visitor-space/components/YearInput/YearInput.tsx
tourlbr: TODO: placeholder is missing

src/modules/visitor-space/components/CreatedFilterForm/CreatedFilterForm.tsx
bertyhell: we'll create a separate jira task for this

bertyhell (approved):

bertyhell:
The bug with year 2023 => 1 jan. 2023 we won't fix, since it is the same issue on avo date filtering as well and nobody has encountered that bug before

src/modules/visitor-space/components/YearInput/YearInput.tsx
bertyhell: placeholder is only visible when year is selected

tourlbr (approved):

tourlbr:
> The bug with year 2023 => 1 jan. 2023 we won't fix, since it is the same issue on avo date filtering as well and nobody has encountered that bug before

@silkederudder-shd In afwezigheid van Ward kan jij deze PR mergen aub?

silkederudder-shd (approved):

PR #626, opened 2023-02-13
Feature/arc 1344

silkederudder-shd:
### Admin
(Ben je leerkracht? item should not be active on these screens, I only added this to demonstrate it)
admin-desktop
admin-mobile

### CP
cp-desktop
cp-mobile-1
cp-mobile-2

### Logged out
logout-desktop
logout-mobile


src/modules/navigation/components/Navigation/Navigation.consts.tsx
```diff
@@ -1,19 +1,26 @@
...
+	NavigationInfo,
+	NavigationPlacement,
+} from '@navigation/services/navigation-service/navigation.types';
 import { Icon, IconName, IconNamesLight } from '@shared/components';
 import { ROUTE_PARTS, ROUTE_PREFIXES, ROUTES } from '@shared/const';
 import { tText } from '@shared/helpers/translate';
 import { Breakpoints } from '@shared/types';
 import { VisitorSpaceInfo } from '@visitor-space/types';
 
+export enum NAVIGATION_DROPDOOWN {
```
tourlbr, line 20: PP: add enum to types file

tourlbr (approved):

tourlbr (approved):

WardVerc1 (approved):

bertyhell:
"Zoeken naar aanbieders" should link to "/bezoek"

image


Going to the search page, crashes the app. You probably need to use ?.replaceAll since the maintainer query param doesn't always have a value

image

bertyhell:
The mobile view of the logged out homepage doesn't look completely correct:
image

bertyhell (changes requested):
see global remarks

silkederudder-shd:
> see global remarks

@bertyhell, see https://meemoo.atlassian.net/browse/ARC-1376. The white line had to be removed on both the home and search page.

silkederudder-shd:
> 

image

bertyhell:
The "bezoek een aanbieder should not be underlined when you're at the homepage:
image

bertyhell (changes requested):
see global remark

src/modules/visitor-space/const/index.tsx
```diff
@@ -53,7 +53,7 @@ export const VISITOR_SPACE_QUERY_PARAM_CONFIG = {
 	// Filters
 	format: withDefault(StringParam, VISITOR_SPACE_QUERY_PARAM_INIT.format),
 	[SEARCH_QUERY_KEY]: ArrayParam,
-	[VisitorSpaceFilterId.Maintainer]: StringParam,
+	[VisitorSpaceFilterId.Maintainer]: withDefault(StringParam, ''),
```
bertyhell, line 56: why is this needed? i would think we just omit the maintainer query param if it is empty?

bertyhell (approved):

PR #642, opened 2023-02-24
fix(translations): use translations in a function and no tHtml in text

bertyhell:
Translations provide 2 functions:
tHtml and tText

tText will output text
tHtml will output a ReactNode with the rendered html

So you cannot use tHtml inside a string

Also the translations need to be fetched from the backend before we can start translating strings, so you always need to fetch translations inside a function and never directly in a const variable on the file level, otherwise it won't correctly be translated.

eg:
```typescript
const myTranslation = tText(....) // wrong
const getMyTranslation = () => tText(...) // correct
```

before:
image

after:
image

### DIFF
--- a/src/modules/visitor-space/components/VisitorSpaceSearchPage/VisitorSpaceSearchPage.tsx
+++ b/src/modules/visitor-space/components/VisitorSpaceSearchPage/VisitorSpaceSearchPage.tsx
@@ -89,14 +89,14 @@ const labelKeys = {
 	search: 'VisitorSpaceSearchPage__search',
 };
 
-function getDefaultOption(): VisitorSpaceDropdownOption {
+const getDefaultOption = (): VisitorSpaceDropdownOption => {
 	return {
 		id: '',
 		label: tText(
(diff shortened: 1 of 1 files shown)


src/modules/visitor-space/components/VisitorSpaceSearchPage/VisitorSpaceSearchPage.tsx
```diff
@@ -89,12 +89,14 @@ const labelKeys = {
...
-	)}`,
-};
+function getDefaultOption(): VisitorSpaceDropdownOption {
+	return {
+		id: '',
+		label: tText(
+			'modules/visitor-space/components/visitor-space-search-page/visitor-space-search-page___pages-bezoekersruimte-publieke-catalogus'
+		),
+	};
+}
```
silkederudder-shd, line 99:
```suggestion
const getDefaultOption = (): VisitorSpaceDropdownOption => {
	return {
		id: '',
		label: tText(
			'modules/visitor-space/components/visitor-space-search-page/visitor-space-search-page___pages-bezoekersruimte-publieke-catalogus'
		),
	};
}
```

WardVerc1 (approved):

PR #647, opened 2023-02-27
fix(ARC-1181): fix scss overwrites

WardVerc1:
See: https://github.com/viaacode/react-components/pull/77


tourlbr (approved):

src/styles/components/_tooltip.scss
```diff
@@ -3,14 +3,14 @@
...
+	background-color: $black !important;
+	border: 1px solid $black !important;
 	display: inline-block;
 	max-width: 19rem !important;
-	padding: 0.5rem;
-	color: $white;
+	padding: 0.5rem !important;
+	color: $white !important;
 	opacity: 0;
-	font-size: 1.3rem;
```
bertyhell, line 13: These styles are being overridden by the avo2-components tooltip which uses the same class: .c-tooltip

A cleaner way of fixing this conflicts is to use a different class in the react-components. eg:
.c-tooltip-component

Then we avoid conflicting css rules alltogether

bertyhell (changes requested):

silkederudder-shd (approved):

PR #648, opened 2023-02-27
fix(ARC-1143): clear year on reset

WardVerc1:


src/modules/visitor-space/components/CreatedFilterForm/CreatedFilterForm.tsx
```diff
@@ -258,6 +258,8 @@ const CreatedFilterForm: FC<CreatedFilterFormProps> = ({ children, className, di
 				values: form,
 				reset: () => {
 					setForm(defaultValues);
```
tourlbr, line 260: PP: null instead of undefined

bertyhell (approved):

silkederudder-shd (approved):

PR #669, opened 2023-03-03
Feature/arc 1264

silkederudder-shd:
**TICKET**
https://meemoo.atlassian.net/browse/ARC-1264

**SCREENSHOT**
https://user-images.githubusercontent.com/77958553/222676143-82611286-143b-45d6-bb17-220b6260088b.mov

**EXTRA INFO**
https://github.com/viaacode/hetarchief-proxy/pull/314#pullrequestreview-1323311154 is needed to work correctly.


WardVerc1 (approved):

src/modules/shared/components/ApproveRequestBlade/ApproveRequestBlade.tsx
```diff
@@ -17,19 +17,27 @@ import {
 	roundToNearestMinutes,
 	startOfDay,
 } from 'date-fns';
+import { isEmpty, isEqual } from 'lodash';
```
bertyhell, line 20: Always import from lodash-es instead of lodash since it is tree shakeable and creates a smaller package size. Could you also fix this in the other locations in the codebase?
```suggestion
import { isEmpty, isEqual } from 'lodash-es';
```

bertyhell (approved):

src/modules/shared/components/ApproveRequestBlade/ApproveRequestBlade.tsx
```diff
@@ -216,6 +278,56 @@ const ApproveRequestBlade: FC<ApproveRequestBladeProps> = (props) => {
...
+						: tText(
+								'modules/cp/components/approve-request-blade/approve-request-blade___er-zijn-x-aantal-mappen-geselecteerd',
+								{
+									count: selectedRefineOptions.length,
+								}
+						  )
+					: tText(
+							'modules/cp/components/approve-request-blade/approve-request-blade___kies-een-map'
+					  )
+			);
```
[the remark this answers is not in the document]
silkederudder-shd: Different options are added because it's designed that way in the Figma file.

src/modules/shared/components/ApproveRequestBlade/ApproveRequestBlade.tsx
silkederudder-shd: Do we still need to lodash package after updating all the imports? Otherwise I'd fully delete it from the project.

bertyhell: i think it is used in the scripts/extract-and-replace-translations.ts script because we don't use webpack to compile that script.

silkederudder-shd: Okay, I'll check it out but will open another PR since it's not related to this ticket (except for the import in this file ofc) ;)

PR #672, opened 2023-03-06
feature/ARC-1312: Add MaterialRequest-blade

WardVerc1:
image
image
image
image

Ticket: https://meemoo.atlassian.net/browse/ARC-1187


src/modules/visitor-space/components/MaterialRequestBlade/MaterialRequestBlade.tsx
```diff
@@ -0,0 +1,240 @@
...
+		try {
+			if (reasonInputValue.length > 0) {
+				const response = await MaterialRequestsService.create({
+					objectId,
+					type: typeSelected,
+					reason: reasonInputValue,
+					requesterCapacity: 'OTHER',
+				});
+				if (response === undefined) {
+					return;
```
bertyhell, line 58: if response is empty something went wrong, show an error toast message, so the user knows something went wrong.
Best to leave the blade open when something went wrong in case they wrote an extensive "reason" in the reason field, so they can copy it, and try again.

bertyhell (approved):

silkederudder-shd (approved):

PR #680, opened 2023-03-08
fix(FilterForm): only set overflow visible for specific filters

silkederudder-shd:
https://meemoo.atlassian.net/browse/ARC-1459

https://user-images.githubusercontent.com/77958553/223671893-71f03a16-b834-4fd7-8a2a-9b58dc578306.mov


src/modules/visitor-space/components/FilterMenu/FilterForm/FilterForm.module.scss
```diff
@@ -27,7 +27,7 @@
 
 .c-filter-form__body {
 	max-height: 100%;
-	overflow: visible;
+	overflow: scroll;
```
bertyhell, line 30: on windows this will always show scrollbars.
image


maybe this is an option?
```suggestion
	overflow-y: auto;
```

bertyhell (approved):

WardVerc1 (approved):

PR #684, opened 2023-03-08
fix(ARC-1175): Added MaterialRequestDetailBlade to beheer screen

WardVerc1:
image

https://meemoo.atlassian.net/browse/ARC-1175
(zie recentste comment)

### DIFF
--- a/src/pages/beheer/materiaalaanvragen/index.tsx
+++ b/src/pages/beheer/materiaalaanvragen/index.tsx
@@ -3,10 +3,11 @@ import clsx from 'clsx';
 import { isEmpty, isNil } from 'lodash';
 import { GetServerSidePropsResult, NextPage } from 'next';
 import { GetServerSidePropsContext } from 'next/types';
-import { ComponentType, ReactNode, useEffect, useMemo, useState } from 'react';
-import { SortingRule, TableState } from 'react-table';
+import { ComponentType, MouseEvent, ReactNode, useEffect, useMemo, useState } from 'react';
+import { Row, SortingRule, TableState } from 'react-table';
 import { useQueryParams } from 'use-query-params';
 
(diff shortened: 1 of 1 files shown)


tourlbr (approved):

src/pages/beheer/materiaalaanvragen/index.tsx
```diff
@@ -121,6 +129,21 @@ const CPMaterialRequestsPage: NextPage<DefaultSeoInfo> = ({ url }) => {
...
+				isOpen={!isLoading && isDetailBladeOpen}
+				onClose={() => setIsDetailBladeOpen(false)}
+				currentMaterialRequestDetail={currentMaterialRequestDetail}
+			/>
+		);
+	};
+
+	const onRowClick = (evt: MouseEvent<HTMLTableRowElement>, row: Row<MaterialRequest>) => {
+		setCurrentMaterialRequest(row.original);
+		setIsDetailBladeOpen(true);
```
bertyhell, line 144: personal preference: this saves state twice.
You can use the setCurrentMaterialRequest to know if the modal should be open. If a material request is selected => open the model, if currentMaterialRequest is null => close the modal

```
const [currentMaterialRequest, setCurrentMaterialRequest] = useState<MaterialRequest | null>(null);
```

Saving state twice sometimes can cause strange behavior if these things get out of sync. eg: if currentMaterialRequest is not set but the isDetailBaldeOpen is set to true.

bertyhell (approved):

src/pages/beheer/materiaalaanvragen/index.tsx
WardVerc1: I'm getting issues with the scrolllock and can't seem to fix this when using 1 state..

PR #708, opened 2023-03-16
Feature/arc 1183 alert popup

BavoVdg:
https://meemoo.atlassian.net/browse/ARC-1276

Needs: 
React-components https://github.com/viaacode/react-components/pull/84 
amdin-core https://github.com/viaacode/react-admin-core-module/pull/136

1. Log in as admin
2. Create a new alert (check dates and make sure it is visible for admins)
3. refresh to refetch alerts
4. profit


src/modules/shared/layouts/AppLayout/AppLayout.tsx
```diff
@@ -60,6 +68,8 @@ import { scrollTo } from '@shared/utils/scroll-to-top';
 
 import packageJson from '../../../../../package.json';
```
tourlbr, line 69: REQ: This is not part of your changes but does not seem to be correct. Please remove

tourlbr (approved):

src/modules/shared/layouts/AppLayout/AppLayout.tsx
BavoVdg: Turns out it is used to set the build version on the window object. We should keep this.

WardVerc1 (approved):

PR #745, opened 2023-03-28
feat(VisitorSpaceSearchPage): use slug instead of or-id

silkederudder-shd:


silkederudder-shd:
> this link still uses the wrong value: image

Should be fixed now ;) 
image

bertyhell (approved):

WardVerc1 (approved):

PR #747, opened 2023-03-28
fix(arc-1554): call export directly on service, without react-query

bertyhell:
https://meemoo.atlassian.net/browse/ARC-1554

I'm not sure this fixes the issue, but i could not reproduce it to begin with. Maybe it only occurs on mac 🍎 🔥 

image


tourlbr (changes requested):
@bertyhell I still have the issue both ways. 10s later the second export works.

bertyhell:
I'll close this PR, since it doesn't fix the issue.
Silke will have a look at this one, since she can reproduce the problem

PR #749, opened 2023-03-28
fix(arc-1497): avoid flashing the login modal after logging in

bertyhell:
https://meemoo.atlassian.net/browse/ARC-1497

You would see the login modal after entering your credentials for a second. Now you should see the page right away.
I also fixed the issue where if you're on the search page as a logged out user, and you login that you would get redirected to the homepage.
Now you'll stay on the search page after login.

image


src/modules/auth/services/auth-service/auth.service.ts
```diff
@@ -22,15 +23,28 @@ export class AuthService {
...
+
+		// Don't redirect the user back to logout, after they logged in
 		if ((originalPath || '').endsWith('/' + ROUTE_PARTS.logout)) {
 			originalPath = '/';
 		}
+
+		// Redirect /slug to the search page with filter
 		if (slug) {
-			originalPath = `/${ROUTE_PARTS.search}/${slug}`;
+			originalPath = `/${ROUTE_PARTS.search}?${ROUTE_PARTS.maintainer}=${slug}`;
```
silkederudder-shd, line 34: I'd use the same enum as used on the filter page to keep them in sync instead of adding those qp to the routes: `VisitorSpaceFilterId.Maintainer`

bertyhell: I'd like to start pulling them apart since VisitorSpaceFilterId is used as the field names in the backend for the api
And ROUTE_PARTS is used in the frontend for everything related to routes. These are also often in dutch. While the api field names will always be in english.

I think we'll have to do that anyways, since meemoo wants to see: `aanbieder=huis-van-alijn` in the url and not `maintainer=huis-van-alijn`

WardVerc1 (approved):

silkederudder-shd (approved):

src/modules/auth/services/auth-service/auth.service.ts
bertyhell: ok, i added a todo to change it
and revert back to `VisitorSpaceFilterId.Maintainer` for now

PR #751, opened 2023-03-28
fix(arc-1566): force month selection arrows to be displayed

bertyhell:
https://meemoo.atlassian.net/browse/ARC-1566


silkederudder-shd:
Looks fixed on the alerts forms but I don't think we always need to add the `flex` styling because it breaks the other datepickers:
image

silkederudder-shd (changes requested):
Looks fixed on the alerts forms but I don't think we always need to add the flex styling because it breaks the styling the other datepickers:
image

PR #756, opened 2023-03-29
Feature/arc 1525

silkederudder-shd:
**Jira ticket**
https://meemoo.atlassian.net/browse/ARC-1525

**Screenrecording**
https://user-images.githubusercontent.com/77958553/228511280-140bb3ac-16ca-4344-93d3-3e7e9946fb5a.mov


src/styles/pages/_object-detail.scss
```diff
@@ -286,6 +298,7 @@ $metadata-collapsed-width: 36vw;
 		.p-object-detail__video {
 			display: grid;
 			grid-template-rows: 1fr auto;
+			margin-top: 0.1rem;
```
silkederudder-shd, line 301: We need this line to prevent this funky thingy:
Screenshot 2023-03-29 at 11 19 51

src/pages/zoeken/[slug]/[ie]/index.tsx
```diff
@@ -735,7 +752,7 @@ const ObjectDetailPage: NextPage<ObjectDetailPageProps> = ({ title, url }) => {
 					<div className="u-pb-24 p-object-detail__actions">
 						<div className="p-object-detail__primary-actions">
 							{showMetadataExportDropdown && renderExportDropdown()}
-							{canRequestMaterial && (
+							{isNotKiosk && (
```
Comment (bertyhell) [inline src/pages/zoeken/[slug]/[ie]/index.tsx:755]: I would leave this as it is, but just as a general rule, we prefer to do user group specific showing/hiding of UI based on permissions and not based on user group.

This is because if meemoo at some point in the future would decide that they do want to show this also for kiosk users, it would require dev changes from us. While if this is checked off by a permission, they could change the permissions in the admin dashboard and the changes would happen automatically without needing dav changes.

We can convert this to a permission in v3

src/styles/pages/_object-detail.scss
bertyhell: Add a comment explaining the issue in the code.

bertyhell (approved):

WardVerc1 (approved):

PR #761, opened 2023-03-31
fix(materialRequest): show organisation if user has organisation

WardVerc1:
image


bertyhell:
for future reference: branches should start with bugfix/, feature/ or release/

silkederudder-shd (approved):

PR #762, opened 2023-03-31
feature(ARC-1230): Delen map

WardVerc1:
image
image
image
image
image
image

Ticket: https://meemoo.atlassian.net/browse/ARC-1230


src/modules/shared/components/ShareFolderBlade/ShareFolderBlade.module.scss
```diff
@@ -0,0 +1,67 @@
...
+
+	&__close-button {
+		margin-bottom: $spacer-sm;
+	}
+
+	&__send-button {
+		background-color: $white;
+		color: $black;
+		border: 0.1rem solid $black !important;
+		font-weight: 600 !important;
```
silkederudder-shd, line 52: According to the Figma file, this should be set to 800 (`$font-weight-bold`)

src/modules/shared/components/ShareFolderBlade/ShareFolderBlade.module.scss
```diff
@@ -0,0 +1,67 @@
...
+		&:hover {
+			border: 0.1rem solid $black;
+			background-color: $black;
+			color: $white;
+		}
+
+		&:disabled {
+			background-color: $zinc;
+			color: $neutral;
+			border: 1px solid $zinc !important;
```
silkederudder-shd, line 64:
```suggestion
			border: .1rem solid $zinc !important;
```

src/modules/shared/components/CopyButton/CopyButton.tsx
```diff
@@ -7,45 +7,61 @@ import { toastService } from '@shared/services/toast-service';
...
-			aria-label={tText(
-				'modules/shared/components/copy-button/copy-button___kopieer-naar-klembord'
-			)}
-			variants="sm"
-			{...props}
-			onClick={clickHandler}
-		/>
-	);
-};
+const CopyButton: FC<ButtonProps & { text: string; enableToast?: boolean; isInputCopy?: boolean }> =
```
silkederudder-shd, line 10: Instead of using the `isInputCopy` boolean to create different buttons, I'd pass the values as properties. This will make this component more flexible and more future proof. 

Since the `CopyButton` has the same props as the `Button` component, you can pass the label, icons, variants,... 

e.g. https://codesandbox.io/s/aged-http-ke5y7p?file=/src/App.js

src/modules/shared/components/ShareFolderBlade/ShareFolderBlade.tsx
```diff
@@ -0,0 +1,161 @@
...
+				<h3 className={styles['c-share-folder-blade__title']}>
+					{tText('pages/account/map-delen/folder-id/index___deel-map')}
+				</h3>
+			)}
+			footer={isOpen && renderFooter()}
+			onClose={handleClose}
+		>
+			<div className={styles['c-share-folder-blade__content']}>
+				<>
+					<dt className={styles['c-share-folder-blade__content-label']}>
```
silkederudder-shd, line 111: `<dt>` and `<dd>` are part of a definition list `<dl>` so if you want to use this correctly, you need to wrap these into `<dl>` (https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dl).

On another note, I don't think it's semantically correct if we use these tags for a form. I think there are better html tags for this such as `label` for the labels. The FormControl already contains an `input` so i think you can wrap those in divs if needed.

silkederudder-shd (changes requested):

src/modules/shared/components/ShareFolderBlade/ShareFolderBlade.tsx
```diff
@@ -0,0 +1,161 @@
...
+				});
+				toastService.notify({
+					maxLines: 3,
+					title: tText('pages/account/map-delen/folder-id/index___map-is-gedeeld'),
+					description: `${tText(
+						'pages/account/map-delen/folder-id/index___map-is-verstuurd-naar'
+					)} ${emailInputValue}`,
+				});
+				handleClose();
+			}
```
bertyhell, line 67: else => show error toast

bertyhell (approved):

silkederudder-shd (approved):

PR #763, opened 2023-03-31
feat(ie): place actions on 1 line

silkederudder-shd:
**Ticket**
https://meemoo.atlassian.net/browse/ARC-1524

**Screenrecording**
https://user-images.githubusercontent.com/77958553/229071032-f4ff0162-f699-4ef2-9e1f-e95b1e52fbd9.mov

### DIFF
--- a/src/pages/zoeken/[slug]/[ie]/index.tsx
+++ b/src/pages/zoeken/[slug]/[ie]/index.tsx
@@ -426,9 +426,9 @@ const ObjectDetailPage: NextPage<ObjectDetailPageProps> = ({ title, url }) => {
 				.replaceAll('{title}', encodeURIComponent(mediaInfo?.name || ''))
 				.replaceAll('{title_serie}', encodeURIComponent(mediaInfo?.series?.[0] || ''));
 			window.open(resolvedFormUrl, '_blank');
-			return;
+		} else {
+			setActiveBlade(MediaActions.RequestMaterial);
 		}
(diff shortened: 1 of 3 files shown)


bertyhell:
according to the ticket, you should also move the buttons to below the title:
image

src/pages/zoeken/[slug]/[ie]/index.tsx
```diff
@@ -425,9 +426,9 @@ const ObjectDetailPage: NextPage<ObjectDetailPageProps> = ({ title, url }) => {
 				.replaceAll('{title}', encodeURIComponent(mediaInfo?.name || ''))
 				.replaceAll('{title_serie}', encodeURIComponent(mediaInfo?.series?.[0] || ''));
 			window.open(resolvedFormUrl, '_blank');
-		} else {
-			setActiveBlade(MediaActions.RequestMaterial);
+			return;
 		}
+		setActiveBlade(MediaActions.RequestMaterial);
```
Comment (bertyhell) [inline src/pages/zoeken/[slug]/[ie]/index.tsx:431]: Personal Preference: i like the original way of writing it more.
if the function has to return something anyway, i would accept your change, but explicitly returning, to then be able to remove the else doesn't really help readability imo.

bertyhell (approved):

src/pages/zoeken/[slug]/[ie]/index.tsx
Comment (silkederudder-shd) [reply src/pages/zoeken/[slug]/[ie]/index.tsx:431]: Okay, I'll put it back to the original format in my next PRs (since I'll be working in the same file).

PR #771, opened 2023-04-03
fix(arc-753): make blade titles smaller

WardVerc1:
image
image
image
image
image
image

Ticket: https://meemoo.atlassian.net/browse/ARC-753


src/modules/account/components/MaterialRequestDetailBlade/MaterialRequestDetailBlade.tsx
```diff
@@ -45,11 +45,11 @@ const MaterialRequestDetailBlade: FC<MaterialRequestDetailBladeProps> = ({
 		<Blade
 			isOpen={isOpen}
 			renderTitle={() => (
-				<h2 className={styles['p-account-my-material-requests__title']}>
+				<h4 className={styles['p-account-my-material-requests__title']}>
```
silkederudder-shd, line 48: Instead of tweaking the HTML outline, add a global classname that sets the same styling as an h4. This way, we don't ruin the outline of the page.

By default, the styling for an h4 has
```
	font-size: $font-size-lg;
	line-height: $line-height-lg;
```

In `Blade.module.scss`, there's already a classname added, you could add the same values there.

silkederudder-shd (approved):

bertyhell:
2 more blades that were missed:
image
image

bertyhell (changes requested):
see global remark

PR #782, opened 2023-04-07
fix(arc-1499): fix no essence flash bug

WardVerc1:
Ticket: https://meemoo.atlassian.net/browse/ARC-1499


src/pages/zoeken/[slug]/[ie]/index.tsx
```diff
@@ -554,13 +553,16 @@ const ObjectDetailPage: NextPage<ObjectDetailPageProps> = ({ title, url }) => {
 		playableUrl: string | undefined,
 		representation: IeObjectRepresentation | undefined
 	): ReactNode => {
+		if (!playableUrl && !mediaInfo) {
+			return <></>;
```
Comment (silkederudder-shd) [inline src/pages/zoeken/[slug]/[ie]/index.tsx:557]: Null is also a reactNode ;) 
```suggestion
			return null;
```

silkederudder-shd (approved):

bertyhell (approved):

PR #783, opened 2023-04-07
fix(arc-1604): fix text search and consultable filters

bertyhell:


WardVerc1:
image
"corner" outside the pill is still shown, it disappears only when clicking outside the searchbar

bertyhell:
@WardVerc1 
> "corner" outside the pill is still shown, it disappears only when clicking outside the searchbar

yes, you're right. This PR only fixes the backend issues with the search. So the reason you don't get any search results.

image

The fix for that text next to the pill when no search results come back is a separate issue.
I would prefer to fix that in a separate PR since this one is getting quite big

WardVerc1 (approved):

silkederudder-shd (approved):

PR #784, opened 2023-04-11
Feature/arc 1190

silkederudder-shd:
**TICKETS**
- Story: https://meemoo.atlassian.net/browse/ARC-1130
- Dev Task: https://meemoo.atlassian.net/browse/ARC-1190

**NOTES**
- Campaign monitor doesn't update the preferences directly in their db, that's why the preference checkbox is, in case you unchecked it, still checked after a refresh. 
- Try to test this with multiple accounts because campaign monitor has a limit of +- 4 post calls per user. After those, you'll receive an error message.

### DIFF
--- a/src/modules/account/const/index.tsx
+++ b/src/modules/account/const/index.tsx
@@ -1,10 +1,19 @@
 import { ReactNode } from 'react';
+import { boolean, object, SchemaOf } from 'yup';
 
+import { CommunicationFormState } from '@account/types';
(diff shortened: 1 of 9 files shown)


WardVerc1 (approved):

bertyhell:
we gaan nog eens moeten kijken naar de backend want die preferences updaten zelfs niet na 40 minuten. Maar dat is een backend issue.

PR #791, opened 2023-04-14
feat(arc-1168): added three clickable tiles content block

WardVerc1:
image
image
image
Ticket: https://meemoo.atlassian.net/browse/ARC-1168
Admin-core PR: https://github.com/viaacode/react-admin-core-module/pull/157


bertyhell:
The block doesn't stay inside the width of the page:
image

I think well either have to make the tiles less wide to make everything fit the page width like in the design. Keeping the tiles square is a bonus.
image

bertyhell (changes requested):
see global remark

silkederudder-shd (approved):

src/styles/admin-core/content-blocks/_three-clickable-tiles.scss
```diff
@@ -0,0 +1,65 @@
+@use "src/styles/abstracts" as *;
+
+$component: "c-block-three-clickable-tiles";
+$tile-size: 27rem;
```
bertyhell, line 4: pages can have different widths, so setting a hard width is not a very good solution:
image
image

Its better to make each tile 33% wide and make the height the same as the width:
```css
.c-block-three-clickable-tiles__term {
    width: 33%;
}
.c-block-three-clickable-tiles__tile-image {
    display: flex;
    justify-content: center;
    align-content: center;
    width: 100%; // change this to 100%
    /* height: 27rem; */
    background-size: cover;
    background-position: center center;
    background-repeat: no-repeat;
    padding: 0 5rem;
    flex-wrap: wrap;
    text-align: center;
    aspect-ratio: 1 / 1; // Add aspect ratio
}
```
image

bertyhell (changes requested):

bertyhell (approved):

PR #798, opened 2023-04-19
fix(arc-1187): fix navigation bar items margin

WardVerc1:
image

Margin between 'Zoeken', 'Bezoek...', 'Over het archief' , ... was 0.

### DIFF
--- a/src/modules/navigation/components/Navigation/Navigation.module.scss
+++ b/src/modules/navigation/components/Navigation/Navigation.module.scss
@@ -215,11 +215,19 @@ $c-navigation-decoration: url("data:image/svg+xml;base64,PHN2ZyB4bWxucz0iaHR0cDo
 		grid-column: 3/4;
 		justify-self: end;
 
-		// Ward: first child element is MaterialRequestCenterButton, which has a blade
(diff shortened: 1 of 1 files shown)


src/modules/navigation/components/Navigation/Navigation.module.scss
```diff
@@ -215,11 +215,19 @@ $c-navigation-decoration: url("data:image/svg+xml;base64,PHN2ZyB4bWxucz0iaHR0cDo
...
+
+			li {
+				// Ward: first child element is MaterialRequestCenterButton, which has a blade
+				// the other navigation items pop up through the blade if the following z-index is removed
+				&:first-child {
+					z-index: get-z-layer("filter");
+				}
+
+				&:nth-child(2) {
+					padding: 0 1.5rem 0 0.5rem;
```
bertyhell, line 229: can't we do this using the grid-gap option instead of doing a padding for one of the items?
eg:
```
.c-navigation__list {
    display: grid;
    grid-auto-flow: column;
    grid-gap: 1.5rem;
}
```

bertyhell (approved):

PR #809, opened 2023-04-21
feat(ARC-1284): Scroll to item when going back

WardVerc1:
https://user-images.githubusercontent.com/113892598/233604874-e95b9e49-d720-46d1-b8ea-71a17df8814b.mov

Ticket: https://meemoo.atlassian.net/browse/ARC-1284


bertyhell:
maybe this is unrelated to the scroll behavior but if i follow these steps, it doesn't scroll:
* as a logged out user
* go to the search page
* scroll down and click on page 3
* scroll down
* click on a tile further down the page => goes to detail page
* use the browser back button

expected behavior: page 3 is loaded, and scrolled down to the item
current behavior: the page 3 is visible in the url for a fraction of a second, but is then reset to page 1. And no scrolling occurs

I assume something on the search page is setting the page to 1, and because of that, the scroll mechanism cannot find the correct element to scroll to. You could try setting breakpoints on all setQuery calls, to see if any of them are being triggered with page: 1.

src/modules/visitor-space/components/VisitorSpaceSearchPage/VisitorSpaceSearchPage.tsx
```diff
@@ -252,6 +258,24 @@ const VisitorSpaceSearchPage: FC = () => {
...
+			setTimeout(() => {
+				const item = document.getElementById(
+					`${lastScrollPosition.itemId}`
+				) as HTMLElement | null;
+
+				item?.scrollIntoView({ block: 'center', behavior: 'smooth' });
+			}, 100);
+			dispatch(setLastScrollPosition({ itemId: '', page: ROUTES.search }));
+		}
+	}, [searchResults?.items]);
```
bertyhell, line 277: shouldn't this array also contain lastScrollPosition?

bertyhell (changes requested):

src/modules/visitor-space/components/VisitorSpaceSearchPage/VisitorSpaceSearchPage.tsx
WardVerc1: image
I get this error than

bertyhell (approved):

silkederudder-shd (approved):

PR #829, opened 2023-04-26
bugfix/ARC-1614: added max width to table cells

WardVerc1:
image
image
image
image

Ticket: https://meemoo.atlassian.net/browse/ARC-1614


silkederudder-shd (approved):

bertyhell:
hmm, this makes all the columns the same width and wastes a lot of space
maybe we should just set a max length on the email adres?
like this:
```
	{
		Header: tText('modules/admin/const/requests___emailadres'),
		accessor: 'visitorMail',
		Cell: ({ row }: VisitRow) => (
			<CopyButton
				className="u-color-neutral u-p-0 c-table__copy"
				icon={undefined}
				variants="text"
				text={row.original.visitorMail}
				title={row.original.visitorMail}
			>
				{truncate(row.original.visitorMail, { length: 35 })}
			</CopyButton>
		),
	},
```

By also setting a title we can ensure the full email is still readable by hovering over it

image

bertyhell (changes requested):
see global remark

bertyhell (approved):

bertyhell:
This pr is replaced with https://github.com/viaacode/hetarchief-client/pull/837 because it has less visual impact. 

This PR:
* sets the width of all columns to the same width
* ensures that the action buttons are always clickable

The other PR https://github.com/viaacode/hetarchief-client/pull/837
* Only limits the width of the email addresses
* has less visual impact
* still has problems with action buttons not being clickable on screensizes between small desktop and phone widths

PR #833, opened 2023-04-27
fix(arc-1573): fix background line coming through overlay

WardVerc1:
image
Ticket: https://meemoo.atlassian.net/browse/ARC-1573

### DIFF
--- a/src/modules/shared/components/Overlay/Overlay.module.scss
+++ b/src/modules/shared/components/Overlay/Overlay.module.scss
@@ -9,7 +9,7 @@ $overlay-opacity-light: 0.4;
 	pointer-events: none;
 	opacity: 0;
 	position: fixed;
-	top: 0;
+	top: -1px;
 	left: 0;
 	right: 0;
 	bottom: 0;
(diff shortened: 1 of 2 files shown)


src/modules/shared/components/Overlay/Overlay.module.scss
```diff
@@ -9,7 +9,7 @@ $overlay-opacity-light: 0.4;
 	pointer-events: none;
 	opacity: 0;
 	position: fixed;
-	top: 0;
+	top: -1px;
```
bertyhell, line 12: wouldn't it be cleaner to change the size of the navigation items to 10rem instead of 9.9rem?
```
@media screen and (min-width: 1400px)
.Navigation_c-navigation__item__ZRI0T {
    margin-bottom: 0.1rem;
    position: relative;
    height: 9.9rem; // change to 10 rem
}
```

bertyhell (changes requested):
see code remark

src/modules/shared/components/Overlay/Overlay.module.scss
silkederudder-shd: ⬆️, negative margins are always a "risk" so if it's possible to fix it with this I'd chose that option.

bertyhell (approved):

silkederudder-shd (approved):

PR #834, opened 2023-04-27
feature/ARC-1291: Added manyResults tile

WardVerc1:
image

Ticket: https://meemoo.atlassian.net/browse/ARC-1291


src/modules/shared/components/MediaCardList/MediaCardList.tsx
```diff
@@ -125,6 +128,26 @@ const MediaCardList: FC<MediaCardListProps> = ({
...
+					title={tText(
+						'modules/shared/components/media-card-list/media-card-list___teveel-resultaten'
+					)}
+					view={view}
+				/>,
+				...tiles.slice(MANY_RESULTS_TILE_POSITION),
+			];
+		}
+		return tiles;
+	};
```
bertyhell, line 149: We always want to show it at the end of the page.
The 10000 is the number of results that elasticsearch will want to output. But even if it is more, we always want to show that tile at the end of the page. So we don't need to do this extra work.
```suggestion
	const insertManyResultsTile = (tiles: ReactNode[]): ReactNode[] => {
		if (showManyResultsTile) {
			tiles.push(<MediaCard
					key="manyResultsTile"
					id="manyResultsTileId"
					type={null}
					preview="/images/more.png"
					title={tText(
						'modules/shared/components/media-card-list/media-card-list___teveel-resultaten'
					)}
					view={view}
				/>)
		}
		return tiles;
	};
```


FYI, if you ever need to insert an element into a specific position in an array, you can use the splice method instead of the slice method: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/splice

src/modules/visitor-space/components/VisitorSpaceSearchPage/VisitorSpaceSearchPage.tsx
```diff
@@ -205,6 +209,11 @@ const VisitorSpaceSearchPage: FC = () => {
 		true
 	);
 
+	// Calculate when to show the manyResultsTile
+	const resultsHasMorethanMaxItems = searchResults && searchResults?.total > RESULTS_MAX;
+	const showManyResultsTile =
+		resultsHasMorethanMaxItems && query.page === PAGE_NUMBER_OF_MANY_RESULTS_TILE;
```
bertyhell, line 215: Doesn't this do the same thing?
```suggestion
	// Calculate when to show the manyResultsTile
	const showManyResultsTile = query.page === PAGE_NUMBER_OF_MANY_RESULTS_TILE;
```

src/modules/shared/components/MediaCardList/MediaCardList.const.ts
```diff
@@ -6,3 +6,8 @@ export const MEDIA_CARD_LIST_GRID_BP_COLS = {
 	[Breakpoints.md]: 2,
 	[Breakpoints.sm]: 1,
 };
+
+export const RESULTS_MAX = 10000;
+export const ITEMS_PER_PAGE = 39;
```
bertyhell, line 11: use the exisiting const instead of redeclaring this one: VISITOR_SPACE_ITEM_COUNT

bertyhell (changes requested):

silkederudder-shd (approved):

PR #847, opened 2023-05-02
feat(ie): always show the navigation bar for kiosks users

silkederudder-shd:
https://meemoo.atlassian.net/browse/ARC-1300?focusedCommentId=40338

Screenshots (before vs after)
Screenshot 2023-05-02 at 14 53 54
Screenshot 2023-05-02 at 14 54 03

Screenshot 2023-05-02 at 14 54 47
Screenshot 2023-05-02 at 14 54 54


WardVerc1 (approved):

src/modules/visitor-space/components/VisitorSpaceNavigation/VisitorSpaceNavigation.tsx
```diff
@@ -111,11 +112,10 @@ const VisitorSpaceNavigation: FC<VisitorSpaceNavigationProps> = ({
...
 
 			<Navigation.Right placement="right">
-				{!isNil(accessEndDate) ? (
+				{!isNil(accessEndDate) && (
 					<span className="u-py-8 u-text-right">{accessEndDate}</span>
-				) : (
-					renderContact()
 				)}
+				{showContactInfo && renderContact()}
 			</Navigation.Right>
```
bertyhell, line 119: OPTIONAL: We always want to render one or the other, never both at the same time
So maybe we should leave the tertiary:
```
			<Navigation.Right placement="right">
				{!isNil(accessEndDate) ? (
					<span className="u-py-8 u-text-right">{accessEndDate}</span>
				) : (
					showContactInfo && renderContact()
				)}
			</Navigation.Right>
```

src/modules/visitor-space/components/VisitorSpaceSearchPage/VisitorSpaceSearchPage.tsx
```diff
@@ -781,7 +782,7 @@ const VisitorSpaceSearchPage: FC = () => {
 	};
 
 	const renderTempAccessLabel = () => {
-		if (isMeemooAdmin || !isPublicCollection) {
+		if (isMeemooAdmin || !isPublicCollection || isKioskUser || isLoggedIn) {
 			// Don't show the temporary access label for:
 			// - MEEMOO admins, since they have access to all visitor spaces
 			// - when a visitor space is selected (https://meemoo.atlassian.net/browse/ARC-1210?focusedCommentId=39708)
```
bertyhell, line 788: not logged in + comments

moving this condition to the start of renderTempAccessLabel is not correct, since we also seem to render the renderResearchWarning for kiosk users in this function

```suggestion
		if (isMeemooAdmin || !isPublicCollection || isKioskUser || !isLoggedIn) {
			// Don't show the temporary access label for:
			// - MEEMOO admins, since they have access to all visitor spaces
			// - when a visitor space is selected (https://meemoo.atlassian.net/browse/ARC-1210?focusedCommentId=39708)
			// When the user is a kiosk user, since kiosk users have unlimited access
			// When the user is not logged in, since then he cannot have temporary access
```

bertyhell (changes requested):

src/modules/visitor-space/components/VisitorSpaceNavigation/VisitorSpaceNavigation.tsx
silkederudder-shd: image
Contact info has been removed from the design. That's why I've added the `showContactInfo` boolean. This way, we don't need to remove the functionality in case it's still needed on other places.

bertyhell (approved):

PR #855, opened 2023-05-03
fix(arc-1213): fix hidden dropdown + reset dropdown

WardVerc1:
image
Ticket: https://meemoo.atlassian.net/browse/ARC-1213

### DIFF
--- a/src/modules/visitor-space/utils/select/options.ts
+++ b/src/modules/visitor-space/utils/select/options.ts
@@ -4,5 +4,11 @@ export const getSelectValue = (
 	options: SelectOption[],
 	optionValue: string | undefined | null
 ): SelectOption | undefined => {
+	if (optionValue === undefined) {
+		return {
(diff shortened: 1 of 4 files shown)


src/modules/visitor-space/utils/select/options.ts
```diff
@@ -4,5 +4,11 @@ export const getSelectValue = (
 	options: SelectOption[],
 	optionValue: string | undefined | null
 ): SelectOption | undefined => {
+	if (optionValue === undefined) {
+		return {
+			label: 'Select...',
```
silkederudder-shd, line 9: Add translation for this.

WardVerc1: image
Can't use useTranslation because it's not a react component.

silkederudder-shd (approved):

bertyhell (approved):

PR #877, opened 2023-05-08
fix(arc-1659): show correct error on objectDetail page

WardVerc1:
image

Ticket: https://meemoo.atlassian.net/browse/ARC-1659
Together with proxy PR: https://github.com/viaacode/hetarchief-proxy/pull/406

### DIFF
--- a/src/pages/zoeken/[slug]/[ie]/index.tsx
+++ b/src/pages/zoeken/[slug]/[ie]/index.tsx
@@ -1334,19 +1334,24 @@ const ObjectDetailPage: NextPage<ObjectDetailPageProps> = ({ title, url }) => {
 		if (mediaInfoIsLoading || visitRequestIsLoading || visitorSpaceIsLoading) {
 			return <Loading fullscreen owner="object detail page: render page content" />;
 		}
+
 		if (mediaInfo) {
 			return <div className="p-object-detail">{renderObjectDetail()}</div>;
 		}
+
 		if (isMediaInfoErrorNoAccess || isMediaInfoErrorNotFound) {
 			if (isErrorSpaceNotFound) {
 				return <ErrorNotFound />;
 			}
-			if (isErrorSpaceNotActive) {
+
+			if (isErrorSpaceNotActive || visitorSpace?.status === VisitorSpaceStatus.Inactive) {
 				return <ErrorSpaceNoLongerActive />;
 			}
+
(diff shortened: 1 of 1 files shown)


src/pages/zoeken/[slug]/[ie]/index.tsx
```diff
@@ -1319,17 +1320,46 @@ const ObjectDetailPage: NextPage<ObjectDetailPageProps> = ({ title, url }) => {
...
 			return <Loading fullscreen owner="object detail page: render page content" />;
 		}
-		if (isErrorSpaceNotActive) {
-			return <ErrorSpaceNoLongerActive />;
+		if (mediaInfo) {
+			return <div className="p-object-detail">{renderObjectDetail()}</div>;
 		}
-		if (isErrorNotFound) {
-			return <ErrorNotFound />;
+		if (isMediaInfoErrorNoAccess || isMediaInfoErrorNotFound) {
```
Comment (silkederudder-shd) [inline src/pages/zoeken/[slug]/[ie]/index.tsx:1340]: Add empty lines between if blocks/statements to make it easier to read / to create a more clear code structure.

silkederudder-shd (changes requested):

src/pages/zoeken/[slug]/[ie]/index.tsx
```diff
@@ -1319,17 +1320,46 @@ const ObjectDetailPage: NextPage<ObjectDetailPageProps> = ({ title, url }) => {
...
+			if (isErrorSpaceNotFound) {
+				return <ErrorNotFound />;
+			}
+			if (isErrorSpaceNotActive) {
+				return <ErrorSpaceNoLongerActive />;
```
Comment (bertyhell) [inline src/pages/zoeken/[slug]/[ie]/index.tsx:1345]: i would expect this case to be shown when i go to:
http://localhost:3200/zoeken/bruzzel/d7d7841f8ad946f8bfa5116b30b9041ccec8fc5521d2404cab08b86c2d41466e231c382d924942de8d586b2f713e85f6

the object id does not exist
and the visitor space has status: inactive

I think this is because the meemoo admin can still fetch inactive spaces, since they need to be able to change their status

So we'll have to change the logic a little bit:
```
if (isErrorSpaceNotActive || visitorSpace?.status === VisitorSpaceStatus.Inactive) {
```

bertyhell (changes requested):

bertyhell (approved):

PR #878, opened 2023-05-08
fix(arc-1650): hide visits with permanent-id on bezoek page

WardVerc1:
image

Ticket: https://meemoo.atlassian.net/browse/ARC-1650

### DIFF
--- a/src/modules/home/components/LoggedInHome/LoggedInHome.tsx
+++ b/src/modules/home/components/LoggedInHome/LoggedInHome.tsx
@@ -113,6 +113,18 @@ const LoggedInHome: FC<DefaultSeoInfo> = ({ url }) => {
 		{ enabled: !!query[VISITOR_SPACE_SLUG_QUERY_KEY], retry: false }
 	);
 
+	// ARC-1650: Do not show all visit accesses for admin users
+	const filterOutPermanentIds = (items?: Visit[]): Visit[] => {
+		if (items === undefined) {
+			return [];
+		}
+		return items.filter((visit) => !visit.id.includes('permanent-id'));
+	};
+
+	const actualActiveVisitRequests = filterOutPermanentIds(active?.items);
+	const actualFutureVisitRequests = filterOutPermanentIds(future?.items);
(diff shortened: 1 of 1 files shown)


src/modules/home/components/LoggedInHome/LoggedInHome.tsx
```diff
@@ -271,7 +271,8 @@ const LoggedInHome: FC<DefaultSeoInfo> = ({ url }) => {
 						</p>
 					</section>
 
-					{(active?.items || []).length > 0 && (
+					{(active?.items.filter((visit) => !visit.id.includes('permanent-id')) || [])
```
silkederudder-shd, line 274: Create a reusable filter function since for this. This way, it'll be easier to make changes. 

Maybe add a comment why we need to do this :)

silkederudder-shd (changes requested):

src/modules/home/components/LoggedInHome/LoggedInHome.tsx
bertyhell, line 274: i would store these arrays as computed values
so you can also use them for the actual generation of the visit space cards

Because now you render all cards if at least one card does not contain a permanent id

But this logic would be better: render only cards that do not have permanent ids, and render nothing if there are none

eg:
```typescript
const actualActiveVisitRequests = active?.items.filter((visit) => !visit.id.includes('permanent-id')) || [];
const actualFutureVisitRequests = future?.items.filter((visit) => !visit.id.includes('permanent-id')) || [];
const actualPendingVisitRequests = pending?.items.filter((visit) => !visit.id.includes('permanent-id')) || [];

{actualActiveVisitRequests.length > 0 && (<section .... actualActiveVisitRequests.map(...) ...)
{actualFutureVisitRequests .length > 0 && (<section .... actualFutureVisitRequests.map(...) ...)
{actualPendingVisitRequests .length > 0 && (<section .... actualPendingVisitRequests.map(...) ...)
```

bertyhell (changes requested):

bertyhell (approved):

silkederudder-shd (approved):

PR #923, opened 2023-05-22
feat(sitemap): add sitemap

silkederudder-shd:
https://meemoo.atlassian.net/browse/ARC-1299

image

Don't forget to add the new environment variable.


src/modules/shared/types/global.d.ts
```diff
@@ -15,6 +15,7 @@ declare global {
 			readonly FLOW_PLAYER_TOKEN: string;
 			readonly FLOW_PLAYER_ID: string;
 			readonly GOOGLE_TAG_MANAGER_ID: string | null;
+			readonly SITEMAP_URL: string;
```
bertyhell, line 18: We should also add the end var to the .env.template file

bertyhell (approved):

WardVerc1 (approved):

PR #995, opened 2023-06-21
fix(arc-1774): add plan een bezoek button to edge case

WardVerc1:
image

Ticket: https://meemoo.atlassian.net/browse/ARC-1774

### DIFF
--- a/src/pages/zoeken/[slug]/[ie]/[name]/index.tsx
+++ b/src/pages/zoeken/[slug]/[ie]/[name]/index.tsx
@@ -340,8 +340,10 @@ const ObjectDetailPage: NextPage<ObjectDetailPageProps> = ({ title, description,
 		isNil(mediaInfo.thumbnailUrl);
 	const showKeyUserPill = mediaInfo?.accessThrough?.includes(IeObjectAccessThrough.SECTOR);
 	const showVisitButton =
-		visitorSpace?.status === VisitorSpaceStatus.Active && canRequestAccess && !isKiosk;
-	const hasContentLicense = mediaInfo?.licenses?.includes(IeObjectLicense.BEZOEKERTOOL_CONTENT);
+		isNil(mediaInfo?.thumbnailUrl) &&
+		mediaInfo?.licenses?.includes(IeObjectLicense.BEZOEKERTOOL_CONTENT) &&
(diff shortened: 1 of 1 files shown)


SHDBramVandenbussche (approved):

src/pages/zoeken/[slug]/[ie]/[name]/index.tsx
```diff
@@ -341,6 +341,7 @@ const ObjectDetailPage: NextPage<ObjectDetailPageProps> = ({ title, description,
 	const showKeyUserPill = mediaInfo?.accessThrough?.includes(IeObjectAccessThrough.SECTOR);
 	const showVisitButton =
 		visitorSpace?.status === VisitorSpaceStatus.Active && canRequestAccess && !isKiosk;
+	const hasContentLicense = mediaInfo?.licenses?.includes(IeObjectLicense.BEZOEKERTOOL_CONTENT);
```
Comment (bertyhell) [inline src/pages/zoeken/[slug]/[ie]/[name]/index.tsx:344]: This check is already part of the existing logic:
```
	const canRequestAccess =
		isNil(
			accessibleVisitorSpaces?.find((space) => space.maintainerId === mediaInfo?.maintainerId)
		) &&
		mediaInfo?.licenses?.includes(IeObjectLicense.BEZOEKERTOOL_CONTENT) &&
		isNil(mediaInfo.thumbnailUrl);
```

The issue is that someone can have access to a limited visitor space. So this check is too strict:
```
isNil(
	accessibleVisitorSpaces?.find((space) => space.maintainerId === mediaInfo?.maintainerId)
)
```

I think we can just omit it, since we already check that there is not thumbnail, so if there is not thumbnail, you don't have access to the content, but the object has content that is accessible.
So this check should do it:
```
	const showVisitButton =
		isNil(mediaInfo.thumbnailUrl) &&
		mediaInfo?.licenses?.includes(IeObjectLicense.BEZOEKERTOOL_CONTENT) &&
		visitorSpace?.status === VisitorSpaceStatus.Active &&
		!isKiosk;		
```
then we can remove the canRequestAccess boolean if it is not used elsewhere.

The current implementation is not correct since it will also show the button for visitor spaces that have an inactive or pending status.

bertyhell (changes requested):

bertyhell (approved):

PR #1003, opened 2023-06-26
bugfix/arc-1834: added filter to hide mapobjects if no keyuser and has license

WardVerc1:
Ticket: https://meemoo.atlassian.net/browse/ARC-1834


bertyhell:
this should be something that is filtered in the backend. Otherwise people can still see the items in the network request in the dev tools
you can add the filtering here: https://github.com/viaacode/hetarchief-proxy/blob/master/src/modules/collections/controllers/collections.controller.ts#L113

bertyhell (changes requested):
see global remark

WardVerc1:
Is handled in proxy: https://github.com/viaacode/hetarchief-proxy/pull/440

PR #1008, opened 2023-06-29
fix(arc-1711): if mobile show tooltip on rightside

WardVerc1:
image

Ticket: https://meemoo.atlassian.net/browse/ARC-1711

### DIFF
--- a/src/modules/shared/components/Pill/Pill.tsx
+++ b/src/modules/shared/components/Pill/Pill.tsx
@@ -2,19 +2,25 @@ import { Tooltip, TooltipContent, TooltipTrigger } from '@meemoo/react-component
 import clsx from 'clsx';
 import { FC, ReactElement } from 'react';
 
+import { useWindowSizeContext } from '@shared/hooks/use-window-size-context';
+import { Breakpoints } from '@shared/types';
+
 import { Icon } from '..';
 
 import styles from './Pill.module.scss';
 import { PillProps } from './Pill.types';
 
 const Pill: FC<PillProps> = ({ className, icon, label, isExpanded }: PillProps): ReactElement => {
+	const windowSize = useWindowSizeContext();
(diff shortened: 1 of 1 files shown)


src/modules/shared/components/Pill/Pill.tsx
```diff
@@ -2,19 +2,25 @@ import { Tooltip, TooltipContent, TooltipTrigger } from '@meemoo/react-component
...
+
 import { Icon } from '..';
 
 import styles from './Pill.module.scss';
 import { PillProps } from './Pill.types';
 
 const Pill: FC<PillProps> = ({ className, icon, label, isExpanded }: PillProps): ReactElement => {
+	const windowSize = useWindowSizeContext();
+	const isMobile = !!(windowSize.width && windowSize.width < Breakpoints.md);
+
```
SHDBramVandenbussche, line 16: `windowSize.width && windowSize.width < Breakpoints.md` will also return a boolean

SHDBramVandenbussche (approved):

src/modules/shared/components/Pill/Pill.tsx
```diff
@@ -2,19 +2,25 @@ import { Tooltip, TooltipContent, TooltipTrigger } from '@meemoo/react-component
...
+	const isMobile = windowSize.width && windowSize.width < Breakpoints.md;
+
 	const rootCls = clsx(className, styles['c-pill'], {
 		[styles['c-pill--expanded']]: isExpanded,
 	});
 
 	const renderTooltipPill = (): ReactElement => (
 		<span className={rootCls}>
-			<Tooltip position="top">
+			<Tooltip position={isMobile ? 'right' : 'top'}>
```
bertyhell, line 23: We could also always show it on the right side that way we don't need another widow width listener
```suggestion
			<Tooltip position="right">
```

bertyhell (approved):

PR #1009, opened 2023-06-29
fix(arc-1717): fix blade issue + faster transition

WardVerc1:
https://github.com/viaacode/hetarchief-client/assets/113892598/60b52621-c9b7-44c3-8e10-732e32574571

Ticket: https://meemoo.atlassian.net/browse/ARC-1717


SHDBramVandenbussche (approved):

bertyhell:
I don't think changing the animation is a good solution for this issue.
It seems to hesitate because either the cursor is over the toast, and when the toast slides under it to close, the cursor isn't on the button anymore, which triggers a hover effect which triggers a show toast again
or it is a react rerender issue, where react rerenders during the close animation


Here are some things we could try in order of best solution to worst solution:

1. update react-toastify package to latest
2. on close => add class to make toast pointer-events: none
3. check if toast is rerendering during the animation. eg: console.log start animation, end animation and put console.log in useEffects(() => {}) without a dependency array, so it triggers on every render
4. add deploy between onClick button and triggering the onClose handler with window.setTimeout(onClose, 500);
5. disable pause on hover: pauseOnHover: 0
6. instantly close the toast if the button is pressed, without an animation. animations are then anly used for self closing toasts.

WardVerc1:
@bertyhell You're right, updated react-toastify fixed the issue ✅👍

bertyhell (approved):

PR #1026, opened 2023-07-19
fix(arc-1522): fix redirects /bezoek/[slug]

WardVerc1:
Ticket: https://meemoo.atlassian.net/browse/ARC-1522


SHDBramVandenbussche (approved):

src/pages/bezoek/[slug]/index.tsx
```diff
@@ -108,6 +118,18 @@ const VisitPage: NextPage<VisitPageProps> = () => {
 	 */
 
 	const renderPageContent = () => {
+		if (showNoAccess) {
```
Comment (bertyhell) [inline src/pages/bezoek/[slug]/index.tsx:121]: isn't it safer to just use the boolean values here, instead of storing them in a useState? Then we're sure they are always updated correctly when isLoggedIn or hasAccess changes
```suggestion
		if (!isLoggedIn || !hasAccess) {
```

bertyhell (approved):

PR #1074, opened 2023-11-02
fix(filter-footer-overflow): reduce filter body maxheight

WardVerc1:
Before:
image

After:
image


src/modules/visitor-space/components/FilterMenu/FilterForm/FilterForm.module.scss
```diff
@@ -30,7 +30,7 @@
 }
 
 .c-filter-form__body {
-	max-height: 45vh;
+	max-height: 40vh;
```
bertyhell, line 33: does this also work if you reduce or increase the chrome window height?
i would have thought this would have to be a fixed value instead of a vh unit

PR #1095, opened 2023-11-20
fix(arc-2001): whole card disabled view when no link

WardVerc1:
image

Ticket: https://meemoo.atlassian.net/browse/ARC-2001


bertyhell:
make the icon also grey: 
image

PR #1097, opened 2023-11-20
Revert: use elasticsearch again to show filter options

bertyhell:
https://meemoo.atlassian.net/browse/ARC-1882

image

image


src/modules/shared/services/api-service/api.service.ts
```diff
@@ -13,7 +13,6 @@ export abstract class ApiService {
 		if (!ApiService.api) {
 			this.api = ky.create({
 				prefixUrl: publicRuntimeConfig.PROXY_URL,
-				timeout: 30000,
```
WardVerc1, line 16: This won't cause issues?

WardVerc1 (approved):

src/modules/shared/services/api-service/api.service.ts
bertyhell: you're right, i added the timeout again

PR #1098, opened 2023-11-20
feat(arc-1798): show error on duplicate material request

WardVerc1:
image

image

Ticket: https://meemoo.atlassian.net/browse/ARC-1798

### DIFF
--- a/src/modules/visitor-space/components/MaterialRequestBlade/MaterialRequestBlade.tsx
+++ b/src/modules/visitor-space/components/MaterialRequestBlade/MaterialRequestBlade.tsx
@@ -1,11 +1,15 @@
-import { Button, RadioButton, TextArea } from '@meemoo/react-components';
+import { Button, OrderDirection, RadioButton, TextArea } from '@meemoo/react-components';
 import clsx from 'clsx';
 import Image from 'next/image';
 import React, { FC, useState } from 'react';
-import { useDispatch } from 'react-redux';
+import { useDispatch, useSelector } from 'react-redux';
 
+import { GroupName } from '@account/const';
+import { selectUser } from '@auth/store/user/user.select';
+import { useGetPendingMaterialRequests } from '@material-requests/hooks/get-pending-material-requests';
 import { MaterialRequestsService } from '@material-requests/services';
 import {
+	MaterialRequestKeys,
 	MaterialRequestObjectType,
 	MaterialRequestRequesterCapacity,
 	MaterialRequestType,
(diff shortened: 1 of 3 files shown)


bertyhell:
like we discussed on the daily, lets move the error so it shows when you try to start the material request process instead of when you submit the blade

src/modules/visitor-space/components/MaterialRequestBlade/MaterialRequestBlade.tsx
```diff
@@ -63,6 +67,15 @@ const MaterialRequestBlade: FC<MaterialRequestBladeProps> = ({
 		type || MaterialRequestType.VIEW
 	);
 	const [reasonInputValue, setReasonInputValue] = useState(reason || '');
+	const user = useSelector(selectUser);
+
+	const { refetch: refetchMaterialRequests } = useGetPendingMaterialRequests(
+		{
+			orderProp: MaterialRequestKeys.maintainer,
+			orderDirection: 'asc' as OrderDirection,
```
bertyhell, line 75:
```suggestion
			orderDirection: OrderDirection.asc,
```

src/pages/zoeken/[slug]/[ie]/[name]/index.tsx
```diff
@@ -502,6 +525,16 @@ const ObjectDetailPage: NextPage<ObjectDetailPageProps> = ({ title, description,
...
+		if (
+			materialRequests?.items &&
+			materialRequests.items.find(
+				(request) => request.objectSchemaIdentifier === mediaInfo?.schemaIdentifier
+			)
+		) {
+			onDuplicateRequest();
+			return;
+		}
+
```
Comment (bertyhell) [inline src/pages/zoeken/[slug]/[ie]/[name]/index.tsx:537]: it might be easier to just fetch the material requests when a user clicks on this button and check if it is in the list
then we don't need all the above logic to keep the list in sync

We usually want to use the react query wrappers so we get loading states, but in this case, we want to check something when the user clicks a button, so an await call directly to the service is simpler

```suggestion
		const materialRequests = await MaterialRequestsService.getAll({
			size: 500,
			isPending: true,
			isPersonal: true,
		});

		if (
			materialRequests?.items &&
			materialRequests.items.find(
				(request) => request.objectSchemaIdentifier === mediaInfo?.schemaIdentifier
			)
		) {
			onDuplicateRequest();
			return;
		}

```

bertyhell (changes requested):

src/modules/visitor-space/components/MaterialRequestBlade/MaterialRequestBlade.tsx
```diff
@@ -101,6 +114,7 @@ const MaterialRequestBlade: FC<MaterialRequestBladeProps> = ({
 				),
 			});
 			onSuccesCreated();
+			refetchMaterialRequests();
```
bertyhell, line 117: do we still need this, since the error is thrown before we even open this blade?

src/modules/visitor-space/components/MaterialRequestBlade/MaterialRequestBlade.tsx
```diff
@@ -63,6 +67,15 @@ const MaterialRequestBlade: FC<MaterialRequestBladeProps> = ({
...
 	const [reasonInputValue, setReasonInputValue] = useState(reason || '');
+	const user = useSelector(selectUser);
+
+	const { refetch: refetchMaterialRequests } = useGetPendingMaterialRequests(
+		{
+			orderProp: MaterialRequestKeys.maintainer,
+			orderDirection: OrderDirection.asc,
+		},
+		{ enabled: !!user && user.groupName !== GroupName.KIOSK_VISITOR }
+	);
```
bertyhell, line 78: do we still need this, since the error is thrown before we even open this blade?

bertyhell (changes requested):

PR #1101, opened 2023-11-21
fix(arc-1960): smaller width so text is in view

WardVerc1:
image

Above is at 350px

image

Ticket: https://meemoo.atlassian.net/browse/ARC-1960


src/styles/admin-core/content-blocks/_content-block-two-columns.scss
```diff
@@ -22,7 +22,7 @@ $component: "c-content-block__rich-text-two-columns";
 			grid-template-columns: repeat(auto-fill, 1fr);
 
 			.o-grid-col-bp2-6 {
-				width: calc(100% - $spacer-1xl) !important;
+				width: calc(90% - $spacer-1xl) !important;
```
bertyhell, line 25: shouldn't we increase the recond value insead of the percentage?
eg:
```suggestion
				width: calc(100% - $spacer-2xl) !important;
```
since we want the text to fill the full width of the phone (100%) minus the size of the margins (spacer px size)

bertyhell (changes requested):

bertyhell (approved):

PR #1103, opened 2023-11-22
feat(arc-2011): add timeout to check user on load

WardVerc1:
(can't take screenshot)

Ticket: https://meemoo.atlassian.net/browse/ARC-2011

### DIFF
--- a/src/modules/shared/layouts/AppLayout/AppLayout.tsx
+++ b/src/modules/shared/layouts/AppLayout/AppLayout.tsx
@@ -149,6 +149,15 @@ const AppLayout: FC<any> = ({ children }) => {
 		return spaces?.items || [];
 	}, [isKioskOrAnonymous, spaces, user]);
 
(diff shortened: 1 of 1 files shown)


src/modules/shared/layouts/AppLayout/AppLayout.tsx
```diff
@@ -149,6 +149,15 @@ const AppLayout: FC<any> = ({ children }) => {
 		return spaces?.items || [];
 	}, [isKioskOrAnonymous, spaces, user]);
 
+	const [isLoaded, setIsloaded] = useState(false);
+
+	useEffect(() => {
+		// ARC-2011: small timeout so login is not shown before user is checked
```
bertyhell, line 155:
```suggestion
		// ARC-2011: small timeout so login is not shown before user is checked
		// If this gives issues in the future, we might want to look into replacing this timeout with selectHasCheckedLogin from the redux store
```

bertyhell (approved):

PR #1119, opened 2024-04-24
fix(AVO-3238): add noindex header if environment is localhost or qas

bertyhell:
https://meemoo.atlassian.net/issues/AVO-3238

image

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -105,6 +105,34 @@ module.exports = withBundleAnalyzer(
 			FLOW_PLAYER_ID: process.env.FLOW_PLAYER_ID,
 			GOOGLE_TAG_MANAGER_ID: process.env.GOOGLE_TAG_MANAGER_ID,
 		},
+		async headers() {
+			if (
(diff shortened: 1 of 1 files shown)


next.config.js
```diff
@@ -105,6 +105,34 @@ module.exports = withBundleAnalyzer(
 			FLOW_PLAYER_ID: process.env.FLOW_PLAYER_ID,
 			GOOGLE_TAG_MANAGER_ID: process.env.GOOGLE_TAG_MANAGER_ID,
 		},
+		async headers() {
+			if (
+				process.env.PROXY_URL.includes('-qas.') ||
+				process.env.PROXY_URL.includes('localhost:3200')
```
ian-emsens-shd, line 111: Optional; I don't think we care about headers in local

ian-emsens-shd (approved):

next.config.js
bertyhell: yeah, it was mostly to be able to see if the header would be there locally. We could remove it, but this way you'll see the same behavior locally as on qas

PR #1120, opened 2024-05-07
feat(content-page-multi-language): fix conflicts multi lang admin core

bertyhell:
requires admin-core pr: https://github.com/viaacode/react-admin-core-module/pull/240


.env.template
```diff
@@ -1,3 +1,9 @@
+# PUBLIC END VARS ######################################################################################################
```
ian-emsens-shd, line 1:
```suggestion
# PUBLIC ENV VARS ######################################################################################################
```

ian-emsens-shd (approved):

PR #1129, opened 2024-05-24
feat(ARC-2076): make all language codes lowercase

bertyhell:
so they match with the nextjs locale


src/modules/content-page/hooks/get-content-page.ts
```diff
@@ -17,6 +18,6 @@ export const useGetContentPageByLanguageAndPath = (
 			throw new Error(`Given path doesn't start with a slash. Received path: ${path}`);
 		}
 
-		return ContentPageService.getContentPageByLanguageAndPath(language, path);
+		return ContentPageService.getContentPageByLanguageAndPath(language as any, path);
```
ian-emsens-shd, line 21: Can we avoid `any`?

ian-emsens-shd (approved):

src/modules/content-page/hooks/get-content-page.ts
bertyhell: no i tried
but the admin-core gets its type from the database: Locale
and the client has its own type: Locale
But for some reason typescript isn't smart enough to figure out that these are the same essential type

PR #1131, opened 2024-05-27
fix(styles): move styles to scss module for detail page

bertyhell:
For some reason these styles were not using the nextjs module scss system. So the styles were lcoated quite far from where they are used.
This PR moves them to the folder where the react component is located
and imports them with the module scss syntax

image


ian-emsens-shd (changes requested):
Not a fan of all these extra `:global`'s, it's originally a css module to avoid that. Is there any other reason to move this single file?

bertyhell:
> Not a fan of all these extra :global's, it's originally a css module to avoid that. Is there any other reason to move this single file?

The goal of module scss is to avoid styles leaking throughout the application. It will add a hash to the selector so it can't leak. Any globals inside a module block, will not be able to leak from its container. So it is much safer then what we had before.

I'm also not a big fan of it, but there is no reason for this file to be loaded as a regular scss file. It should have been programmed originally like i did it now.

PR #1132, opened 2024-05-28
feat(ARC-2141): rework styling

RobbeBierebeeck:


src/styles/admin-core/content-blocks/_enclosed-content.scss
```diff
@@ -41,52 +38,54 @@ $component: "c-block-enclosed-content";
 		&__card {
 			a {
 				text-decoration: none;
+				width: 100%;
 			}
 
+			background-color: $white;
+			border: 1px solid #adadad66;
```
bertyhell, line 45: don't use colors directly
we should always use color variables or create a variable if it doesn't exist

bertyhell (approved):

PR #1133, opened 2024-05-28
feat(ARC-2197): enable server side rendering for homepage content page

bertyhell:

admin-core pr: https://github.com/viaacode/react-admin-core-module/pull/255

image

Fixes to get server side rendering working:
* Make admin-core content page info's serializable (DbContentPage vs ContentPageInfo)
* use isomorpic dompurify package so it runs on the server as well as in the browser
* init admin-core-config as soon as possible, so we don't have to wait for it during server side rendering
* replace instances of window.location.origin with env variable CLIENT_URL
* add shortcuts to hooks: if serverside rendering => user is not logged in
* renamed old LoggedInHome and LoggedOutHome to LoggedInVisitorSpacesHome and LoggedOutVisitorSpacesHome
* add react-query prefetch calls and pass the dehydraded state to the components


src/modules/auth/wrappers/with-auth/with-auth.tsx
```diff
@@ -62,7 +60,6 @@ export const withAuth = (
 					await toHome();
 				} else {
 					// User is logged out, but the page can be seen without logging in
-					setShowPage(true);
 				}
```
ian-emsens-shd, line 63: Intentional empty `else`?

ian-emsens-shd (approved):

src/modules/auth/wrappers/with-auth/with-auth.tsx
bertyhell: yes, the comment describe why nothing should happen here
the minifier will probably remove the empty else

bertyhell:
merged in separate branch: release/sprint-2

PR #1136, opened 2024-05-29
feat(ARC-2197): fix server side rendering of the /slug page, 404 page

bertyhell:

image
image

image
image

image
image

### DIFF
--- a/src/modules/admin/wrappers/admin-core-config.tsx
+++ b/src/modules/admin/wrappers/admin-core-config.tsx
@@ -203,6 +203,7 @@ export function getAdminCoreConfig(router: NextRouter | null, locale: Locale): A
 			},
 		},
 		routes: ADMIN_CORE_ROUTES_BY_LOCALE[locale],
+		locale: locale as any,
 		env: {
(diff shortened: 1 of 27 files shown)


src/modules/admin/wrappers/admin-core-config.tsx
```diff
@@ -203,6 +203,7 @@ export function getAdminCoreConfig(router: NextRouter | null, locale: Locale): A
 			},
 		},
 		routes: ADMIN_CORE_ROUTES_BY_LOCALE[locale],
+		locale: locale as any,
```
ian-emsens-shd, line 206: Avoid `any` if possible

ian-emsens-shd (approved):

src/modules/admin/wrappers/admin-core-config.tsx
bertyhell: yeah, i tried, but the Locale from admin-core doesn't seem to be compatible with the Locale enum from the client :s

bertyhell:
merged in release/sprint-2

PR #1137, opened 2024-05-29
Fix server side rendering issues with search page filter controls

bertyhell:

image
image

### DIFF
--- a/src/modules/shared/components/Html/Html.tsx
+++ b/src/modules/shared/components/Html/Html.tsx
@@ -15,7 +15,7 @@ export interface HtmlProps {
 const Html: FunctionComponent<HtmlProps> = ({
 	content,
 	sanitizePreset = RICH_TEXT_SANITIZATION,
-	type = 'p',
+	type = 'div',
 	className,
 }) => {
 	const Type = type;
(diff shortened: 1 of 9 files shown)


src/modules/shared/components/Html/Html.tsx
```diff
@@ -15,7 +15,7 @@ export interface HtmlProps {
 const Html: FunctionComponent<HtmlProps> = ({
 	content,
 	sanitizePreset = RICH_TEXT_SANITIZATION,
-	type = 'p',
+	type = 'div',
```
ian-emsens-shd, line 18: Make sure to check styles of components that implement `Html` for any selectors relying on the previous `p`-tag default.

src/modules/visitor-space/components/NoServerSideRendering/NoServerSideRendering.tsx
```diff
@@ -0,0 +1,15 @@
+import { FC, ReactNode, useEffect, useState } from 'react';
+
+interface NoSsrProps {
+	children: ReactNode;
+}
+
+export const NoServerSideRendering: FC<NoSsrProps> = ({ children }) => {
```
ian-emsens-shd, line 7: Does this component replace `import { isServerSideRendering } from '@shared/utils'` entirely?
If so, probably should take this opportunity to deprecate / refactor instances of `isServerSideRendering`.

ian-emsens-shd (approved):

src/modules/visitor-space/components/NoServerSideRendering/NoServerSideRendering.tsx
bertyhell: no, NoServerSideRendering only for rendering things
the isServerSideRendering() is still usefull for switching in logic not related to rendering

bertyhell:
merged in release/sprint-2

PR #1142, opened 2024-06-04
ARC-2113 Cookbiebot multilanguage

bertyhell:
make cookiebot banner and cookie declaration follow the selected language

image
image


src/modules/cookie-policy/CookiePolicy.tsx
```diff
@@ -31,10 +41,16 @@ export const CookiePolicy: FC<DefaultSeoInfo> = ({ url }) => {
...
+		script.src = stringifyUrl({
+			url: 'https://consent.cookiebot.com/e17bca33-78a0-484e-a204-e05274a65598/cdreport.js',
+			query: {
+				referer: 'hetarchief.be',
+				culture: locale,
+			},
+		});
 		document.head.appendChild(script);
-	}, []);
+	}, [locale]);
```
ian-emsens-shd, line 53: Split this into two hooks, one that inits a single script tag, the other handling locale changes.
Doing so will avoid an extra script tag (with a duplicate `id`) being added to `document.head` every time the user swaps language.

ian-emsens-shd (changes requested):

src/modules/cookie-policy/CookiePolicy.tsx
bertyhell: the script tag source contains the locale
so it has to create a new script evety time
unless you're saying updating the source of the exisiting script tag, but i'm not sure that would reload the script

i think its safer like i do now: 
* delete the script tag by id
* add a script tag with the correct locale

if the locale changes
* delete the script tag by id
* add a script tag with the correct locale

ian-emsens-shd (approved):

PR #1146, opened 2024-06-07
fix(ARC-2117): fix locale fetching in notification services

bertyhell:
part of a PR group:
* https://github.com/viaacode/react-admin-core-module/pull/258
* https://github.com/viaacode/hetarchief-client/pull/1146
* https://github.com/viaacode/hetarchief-proxy/pull/479

image
image


src/modules/shared/services/translation-service/translation.service.ts
```diff
@@ -9,4 +15,19 @@ export class TranslationService {
 
 		return response ?? {};
 	}
+
+	public static getLocale(): Locale {
+		if (isServerSideRendering()) {
+			return Locale.nl; // Window not available in server side rendering
+		}
```
ian-emsens-shd, line 22: To my understanding this will make only the Dutch-translated version indexable by crawlers, am I missing something?

ian-emsens-shd (changes requested):

PR #1152, opened 2024-06-14
feat(ARC-2223): add font-weight to ctas on blades

RobbeBierebeeck:


src/modules/navigation/components/PersonalInfoBlade/PersonalInfoBlade.module.scss
```diff
@@ -43,5 +43,9 @@ $component: "c-personal-info-blade";
 
 	&__send-button {
 		margin-bottom: $spacer-sm;
+
+		div {
+			font-weight: 600;
```
bertyhell, line 48: use scss variables where possible:
```suggestion
			font-weight: $font-weight-bold;
```

bertyhell (approved):

PR #1155, opened 2024-06-17
feat(ARC-2230): style the thumbnail strip and back button

bertyhell:
https://meemoo.atlassian.net/browse/ARC-2230

image
image
image
image

### DIFF
--- a/src/modules/iiif-viewer/IiifViewer.tsx
+++ b/src/modules/iiif-viewer/IiifViewer.tsx
@@ -1,7 +1,4 @@
 import { Button } from '@meemoo/react-components';
-import clsx from 'clsx';
-import { clamp } from 'lodash-es';
-import React, { forwardRef, useCallback, useEffect, useImperativeHandle, useState } from 'react';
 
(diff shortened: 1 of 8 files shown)


src/modules/iiif-viewer/IiifViewer.tsx
```diff
@@ -370,7 +373,7 @@ const IiifViewer = forwardRef<IiifViewerFunctions, IiifViewerProps>(
 							</button>
 						);
 					})}
-				</div>
```
ian-emsens-shd, line 373: Add comment why a custom scrollbar is used

ian-emsens-shd (approved):

src/modules/iiif-viewer/IiifViewer.tsx
bertyhell: ok, i've added:
			// Use custom scrollbar because on windows the default scrollbar is big and ugly
			// and here it shows up on the right side of the reference strip, so it extra noticeable

bertyhell:

PR #1157, opened 2024-06-17
feat(ARC-2238): fix grid view for iiif viewer on mobile and desktop

bertyhell:
https://meemoo.atlassian.net/browse/ARC-2238

image
image

image

### DIFF
--- a/src/modules/ie-objects/ObjectDetailPage.tsx
+++ b/src/modules/ie-objects/ObjectDetailPage.tsx
@@ -1413,7 +1413,11 @@ export const ObjectDetailPage: FC<DefaultSeoInfo> = ({ title, description, image
 	};
 
 	const renderOcrBlade = () => {
-		if (!isOpenSeaDragonPoc || !OPEN_SEA_DRAGON_POC_IMAGE_INFOS[activeImageIndex].altoUrl) {
+		if (
+			!isOpenSeaDragonPoc ||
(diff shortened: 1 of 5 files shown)


src/modules/ie-objects/ObjectDetailPage.tsx
```diff
@@ -1413,7 +1413,11 @@ export const ObjectDetailPage: FC<DefaultSeoInfo> = ({ title, description, image
 	};
 
 	const renderOcrBlade = () => {
-		if (!isOpenSeaDragonPoc || !OPEN_SEA_DRAGON_POC_IMAGE_INFOS[activeImageIndex].altoUrl) {
+		if (
+			!isOpenSeaDragonPoc ||
+			!OPEN_SEA_DRAGON_POC_IMAGE_INFOS[activeImageIndex].altoUrl ||
+			(activeTab === ObjectDetailTabs.Media && isMobile)
```
ian-emsens-shd, line 1419: Hide elements responsively using css, not js. This keeps display rules centralised and avoids tearing your hair out during debugging

ian-emsens-shd (approved):

src/modules/ie-objects/ObjectDetailPage.tsx
bertyhell: This ocr drawer will be moved according to "yet to be approved design"
then it will be inline in de metadata sidebar, so this will become a non issue

i will leave it as is, until the new design is approved, then make sure the isMobile variable isn't used for this anymore

image

bertyhell:

PR #1158, opened 2024-06-17
feat(ARC-2244): highlight search terms in ocr text

bertyhell:
https://meemoo.atlassian.net/browse/ARC-2244

image
image

### DIFF
--- a/src/modules/ie-objects/ObjectDetailPage.tsx
+++ b/src/modules/ie-objects/ObjectDetailPage.tsx
@@ -802,19 +802,22 @@ export const ObjectDetailPage: FC<DefaultSeoInfo> = ({ title, description, image
 		[metadataExportDropdownOpen, onExportClick, tHtml, tText]
 	);
 
-	const highlighted = (toHighlight: string) => (
-		<Highlighter
-			searchWords={
-				query[QUERY_PARAM_KEY.HIGHLIGHTED_SEARCH_TERMS]
-					? decodeURIComponent(
-							query[QUERY_PARAM_KEY.HIGHLIGHTED_SEARCH_TERMS] as string
-					  ).split(HIGHLIGHTED_SEARCH_TERMS_SEPARATOR)
-					: []
-			}
-			autoEscape={true}
-			textToHighlight={toHighlight}
-		/>
-	);
+	const highlighted = (toHighlight: string, searchTerms?: string[]) => {
+		const searchWords = searchTerms
+			? searchTerms
+			: query[QUERY_PARAM_KEY.HIGHLIGHTED_SEARCH_TERMS]
+			? decodeURIComponent(query[QUERY_PARAM_KEY.HIGHLIGHTED_SEARCH_TERMS] as string).split(
+					HIGHLIGHTED_SEARCH_TERMS_SEPARATOR
+			  )
+			: [];
+		return (
+			<Highlighter
+				searchWords={searchWords}
+				autoEscape={true}
+				textToHighlight={toHighlight}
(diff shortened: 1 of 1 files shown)


src/modules/ie-objects/ObjectDetailPage.tsx
```diff
@@ -802,19 +802,22 @@ export const ObjectDetailPage: FC<DefaultSeoInfo> = ({ title, description, image
...
-		/>
-	);
+	const highlighted = (toHighlight: string, searchTerms?: string[]) => {
+		const searchWords = searchTerms
+			? searchTerms
+			: query[QUERY_PARAM_KEY.HIGHLIGHTED_SEARCH_TERMS]
+			? decodeURIComponent(query[QUERY_PARAM_KEY.HIGHLIGHTED_SEARCH_TERMS] as string).split(
+					HIGHLIGHTED_SEARCH_TERMS_SEPARATOR
+			  )
+			: [];
```
ian-emsens-shd, line 812: Note that `|| []` does not make `!searchTerms` redundant

```suggestion
		const param = query[QUERY_PARAM_KEY.HIGHLIGHTED_SEARCH_TERMS];
		let searchWords = searchTerms || [];

		if (!searchTerms && param) {
			searchWords = decodeURIComponent(param).split(HIGHLIGHTED_SEARCH_TERMS_SEPARATOR);
		}
```

src/modules/ie-objects/ObjectDetailPage.tsx
```diff
@@ -1487,11 +1490,10 @@ export const ObjectDetailPage: FC<DefaultSeoInfo> = ({ title, description, image
 									setIsOcrEnabled(!isOcrEnabled);
 								}}
 							>
-								<Highlighter
-									searchWords={searchOcrText.split(' ')}
-									autoEscape={true}
-									textToHighlight={textLocation.text}
-								/>
+								{highlighted(
```
ian-emsens-shd, line 1493: This should not be a function, it always outputs the same component.

ian-emsens-shd (changes requested):

src/modules/ie-objects/ObjectDetailPage.tsx
bertyhell: i'm not sure i follow what you're trying to say

the code should be as follows:
* use searchTerms if they exist
* use query[QUERY_PARAM_KEY.HIGHLIGHTED_SEARCH_TERMS] if they exist
* use [] if all else fails

Maybe it's clearer if i write it using if else instead of a double tertiary operator:
```typescript
		let searchWords: string[];
		if (searchTerms) {
			searchWords = searchTerms;
		} else if (query[QUERY_PARAM_KEY.HIGHLIGHTED_SEARCH_TERMS]) {
			searchWords = decodeURIComponent(query[QUERY_PARAM_KEY.HIGHLIGHTED_SEARCH_TERMS] as string).split(
				HIGHLIGHTED_SEARCH_TERMS_SEPARATOR
			);
		} else {
			searchWords = [];
		}
```

src/modules/ie-objects/ObjectDetailPage.tsx
bertyhell: nice!
finally a "split this into a component" remark that we can agree on 😄 

i've moved it to a separate component <HighlightSearchTerms toHighlight={text} />

bertyhell:
fixed remarks in https://github.com/viaacode/hetarchief-client/pull/1160

bertyhell:

PR #1163, opened 2024-06-19
feat: add checkbox to requestAccessblade

RobbeBierebeeck:


src/modules/home/components/RequestAccessBlade/RequestAccessBlade.tsx
```diff
@@ -21,14 +25,22 @@ const labelKeys: Record<keyof RequestAccessFormState, string> = {
...
 	const { data: visitorSpace } = useGetVisitorSpace(
 		query[QUERY_PARAM_KEY.VISITOR_SPACE_SLUG_QUERY_KEY] || null
 	);
+	const { data: preferences } = useGetNewsletterPreferences(user?.email);
+
 	const [isSubmitting, setIsSubmitting] = useState<boolean>(false);
 
+	const [isSubscribedToNewsletter, setIsSubscribedToNewsletter] = useState<boolean>(
+		preferences?.newsletter || false
+	);
```
bertyhell, line 42: is it better to default to false or to true here?

case 1A:
* someone is subscribed already + default to false
* the page loads
* preferences has not been fetched yet
* so this value defaults to false
* the checkbox is shown
* the preferences finish loading
* the value becomes true
* the checkbox remains hidden (which is strange behavior)

case 1B:
* someone is not subscribed yet + default to false
* the page loads
* preferences has not been fetched yet
* so this value defaults to false
* the checkbox is shown
* the preferences finish loading
* the value becomes true
* the checkbox remains visible

case 2A:
* someone is subscribed already + default to true
* the page loads
* preferences has not been fetched yet
* so this value defaults to true
* the checkbox is hidden
* the preferences finish loading
* the value becomes true
* the checkbox remains hidden

case 2B:
* someone is not subscribed yet + default to true
* the page loads
* preferences has not been fetched yet
* so this value defaults to true
* the checkbox is hidden
* the preferences finish loading
* the value becomes true
* the checkbox becomes visible


So the question is: what is worse:
* a checkbox that is visible for a second and then dissapears
* or a checkbox that is hidden and then shows up

i think case 2 is the better of the 2

bertyhell (changes requested):

src/modules/home/components/RequestAccessBlade/RequestAccessBlade.tsx
```diff
@@ -53,8 +79,19 @@ const RequestAccessBlade: FC<RequestAccessBladeProps> = ({ onSubmit, isOpen, ...
...
+						checked={isSubscribedToNewsletter}
+						label={tHtml(
+							'modules/home/components/request-access-blade/request-access-blade___schrijf-je-in-voor-de-nieuwsbrief'
+						)}
+						onClick={() => setIsSubscribedToNewsletter((prevState) => !prevState)}
+					/>
+				) : null}
 				<FormControl
-					className="u-mb-8 u-mb-24:md"
+					className="u-mx-8 u-mb-24:md"
```
ian-emsens-shd, line 94: Plakken de errors nu niet tegen dit veld?

ian-emsens-shd (approved):

PR #1165, opened 2024-06-20
fix(ARC-2187): empty character is placed when crumb is empty and brea…

RobbeBierebeeck:
…dcrumbs are centered


ian-emsens-shd (approved):

bertyhell (approved):

bertyhell:
try to keep the commit messages under 50 characters
otherwise github splits them into title and body

PR #1168, opened 2024-06-28
feat(ARC-2188): remove meemoo id from object detail

RobbeBierebeeck:


src/modules/ie-objects/ie-objects.consts.tsx
```diff
@@ -354,10 +354,6 @@ export const METADATA_FIELDS = (mediaInfo: IeObject): MetadataItem[] => [
 		title: tText('modules/ie-objects/const/index___oorsprong'),
 		data: mediaInfo.meemooOriginalCp,
 	},
-	{
-		title: tText('modules/ie-objects/const/index___meemoo-identifier'),
-		data: mediaInfo.meemooIdentifier,
```
bertyhell, line 359: i would also delete this meemooIdentifier from all the types

eg:
* https://github.com/viaacode/hetarchief-client/blob/bd956bb9b9ef7e8e3d692f6606132d95f78868f7/src/modules/account/types/collections.ts#L21
* https://github.com/viaacode/hetarchief-client/blob/bd956bb9b9ef7e8e3d692f6606132d95f78868f7/src/modules/account/views/MyFolders.tsx#L416

The only place we cannot fix it yet is around getRelatedObjects. This will be replaced by a new query that uses the schemaIdentifier to find the related objects. So for now you can also replace it with the schemaIdentifier:
https://github.com/viaacode/hetarchief-client/blob/0226dff4da9ea787a2e3f54280d23811cc34e59f/src/modules/ie-objects/ObjectDetailPage.tsx#L311-L316

Locations where it is shown next to the schemaIdentifier, we can delete
Locations where it is shown without the schemaIdentifier we can replace it with the schemaIdentifier

Currently the schema identifier is very long, but in the v3 objects this identfier will become shorter similar to the format of the v2 meemooIdentifier.

bertyhell (changes requested):

PR #1176, opened 2024-07-17
ARC-2416 Fix build INT/PRD

spacid:
Every environment should be build using specific parameters. This is to ensure the server-side rendering uses the correct information when pre-rendering components.

Pass correct param for INT/PRD


spacid:
Closed because of branch rename to include the JIRA ticket number.

PR #1184, opened 2024-07-26
feat(ARC-2268): update folder share endpoint

RobbeBierebeeck:


src/modules/shared/components/ShareFolderBlade/ShareFolderBlade.tsx
```diff
@@ -55,17 +54,10 @@ const ShareFolderBlade: FC<ShareFolderBladeProps> = ({ isOpen, onClose, folderId
...
 						to: emailInputValue,
 						consentToTrack: 'unchanged',
-						data: {
-							sharer_email: user.email,
-							sharer_name: user.fullName,
-							folder_name: folderName,
-							folder_sharelink: link,
-						},
 					},
 				});
```
bertyhell, line 62: This is a bit of a strange signature for this function
why isn't it just:
```
await foldersService.shareFolder(folderId, emailInputValue);
```
since the consertToTrack isn't needed anymore since that is part of the internal logic of the endpoint in the backend

src/modules/ie-objects/ObjectDetailPage.tsx
```diff
@@ -725,15 +741,23 @@ export const ObjectDetailPage: FC<DefaultSeoInfo> = ({ title, description, image
...
-					window.open(
-						`${publicRuntimeConfig.PROXY_URL}/${NEWSPAPERS_SERVICE_BASE_URL}/${ieObjectId}/${IE_OBJECTS_SERVICE_EXPORT}/zip`
-					);
+					setCopyrightModalOpen(true);
+					setOnConfirmCopyright(() => () => {
+						window.open(
+							`${publicRuntimeConfig.PROXY_URL}/${NEWSPAPERS_SERVICE_BASE_URL}/${ieObjectId}/${IE_OBJECTS_SERVICE_EXPORT}/zip`
+						);
+						handleOnDownloadEvent();
+					});
```
bertyhell, line 750: OPTIONAL: preferably we would have:
* user clicks button => handleExportButtonClicked
* exportButtonClicked => setConfirmModalOpen(true)
* modal is shown
* modal has confirmCallback => handleExport()
* handleExport does the window.open(exportLink) + trigger event

bertyhell (approved):

PR #1186, opened 2024-07-29
feat(ARC-2245): add ocr metadata to detail page from alto xml

bertyhell:
https://meemoo.atlassian.net/browse/ARC-2245

image
image

### DIFF
--- a/src/modules/ie-objects/ie-objects.consts.tsx
+++ b/src/modules/ie-objects/ie-objects.consts.tsx
@@ -20,6 +20,7 @@ import {
 	mapBooleanToMetadataData,
 	mapObjectToMetadata,
 } from '@ie-objects/utils/map-metadata';
+import type { SimplifiedAlto } from '@iiif-viewer/IiifViewer.types';
 import { Icon } from '@shared/components/Icon';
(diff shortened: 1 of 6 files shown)


src/modules/ie-objects/ie-objects.consts.tsx
```diff
@@ -377,8 +378,54 @@ export const MEDIA_ACTIONS = ({
  * Metadata
  */
 
+const getOcrMetadataFields = (simplifiedAlto: SimplifiedAlto | null): MetadataItem[] => {
+	const metadataFields: MetadataItem[] = [];
+	if (simplifiedAlto) {
+		if (simplifiedAlto.description.softwareName) {
```
RobbeBierebeeck, line 384: Isn't it better to use a switch case in this situation?

RobbeBierebeeck (approved):

src/modules/ie-objects/ie-objects.consts.tsx
bertyhell: no cause each case should be executed. If i only wanted to add one a switch would be the better option

PR #1213, opened 2024-08-23
fix(ARC-2235): Add highlight current search term on newspaper essence

RobbeBierebeeck:
…ighlight

fix(ARC-2235): Add highlight current search term on newspaper essence

### DIFF
--- a/src/modules/ie-objects/hooks/get-ie-objects-info.ts
+++ b/src/modules/ie-objects/hooks/get-ie-objects-info.ts
@@ -1,4 +1,5 @@
 import { type QueryClient, useQuery, type UseQueryResult } from '@tanstack/react-query';
+import { useRouter } from 'next/router';
 
 import { QUERY_KEYS } from '@shared/const/query-keys';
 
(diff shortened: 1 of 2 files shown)


src/modules/ie-objects/hooks/get-ie-objects-info.ts
```diff
@@ -16,11 +17,28 @@ export const useGetIeObjectInfo = (
...
+	return useQuery(
+		[QUERY_KEYS.getIeObjectsInfo, id],
+		async () => {
+			let newId;
+			if (id.length > 36) {
+				const v3Identifier = await IeObjectsService.schemaIdentifierLookup(id);
+				newId = v3Identifier.schemaIdentifierV3;
+				const url = new URL(window.location.href);
+				url.pathname = url.pathname.replace(id, newId);
+				await router.replace(url.toString());
```
bertyhell, line 30: it might be better to move this code to the detail page
since this hook should only get some info from the api, and not have any other side effects

imagine we want to get an ie object somewhere else in the code, we might not want to do the redirect there.

something like:
```typescript
useEffect(() => {
	if (router.query.id.length > MIN_LENGTH_SCHEMA_IDENTIFIER_V2 && router.query.id !== mediaItem.schema_identifier) {
		const url = router.asPath;
		url= url.replace('/' + id + '/', '/' + newId + '/');
		await router.replace(url);
	}
}, [mediaItem]);
```

src/modules/ie-objects/hooks/get-ie-objects-info.ts
```diff
@@ -16,11 +17,28 @@ export const useGetIeObjectInfo = (
...
-		...options,
-	});
+	const router = useRouter();
+	return useQuery(
+		[QUERY_KEYS.getIeObjectsInfo, id],
+		async () => {
+			let newId;
+			if (id.length > 36) {
+				const v3Identifier = await IeObjectsService.schemaIdentifierLookup(id);
+				newId = v3Identifier.schemaIdentifierV3;
```
bertyhell, line 27: When i read "v3identifier" i expect it to contain a string. This naming is better:
```suggestion
				const v3IdentifierResponse = await IeObjectsService.schemaIdentifierLookup(id);
				newId = v3IdentifierResponse.schemaIdentifierV3;
```

bertyhell (approved):

bertyhell (approved):

PR #1214, opened 2024-08-23
fix(ARC-2391): add language to logo url

RobbeBierebeeck:


src/modules/shared/layouts/AppLayout/AppLayout.tsx
```diff
@@ -75,7 +75,7 @@ import { useGetAllActiveVisits } from '@visit-requests/hooks/get-all-active-visi
 
 import packageJson from '../../../../../package.json';
 
-import styles from './AppLayout.module.scss';
+import styles from './AppLayout.module.scss'; // We want to make sure config gets fetched here, no sure why anymore
```
bertyhell, line 78: I don't think this comment makes sense here?

src/modules/shared/layouts/AppLayout/AppLayout.tsx
```diff
@@ -332,7 +332,7 @@ const AppLayout: FC<any> = ({ children }) => {
 				node: (
 					<div
 						onClick={() => {
-							window.open(window.location.origin, '_self');
+							window.open(`${window.location.origin}/${locale}`, '_self');
```
bertyhell, line 335: We could use a link for this, so we don't have to reload the complete react app:
It also shows the link you will navigate to at the bottom of the screen when we use a link:
image

```typescript
				node: (
					<Link href={'/' + locale} passHref>
						<a className="c-navigation__logo--hamburger">
							<HetArchiefLogo
								className="c-navigation__logo c-navigation__logo--list"
								type={isMobile ? HetArchiefLogoType.Dark : HetArchiefLogoType.Light}
							/>
						</a>
					</Link>
				),
```

bertyhell (changes requested):

bertyhell (approved):

PR #1227, opened 2024-09-06
ARC-2171 Autocomplete filter reusable form + 3 extra filters

bertyhell:
https://meemoo.atlassian.net/browse/ARC-2171

Make the autocomplete filter form reusable for
* newspaperSeriesName
* locationCreated
* mentions

proxy PR: https://github.com/viaacode/hetarchief-proxy/pull/512

image
image

image
image

image
image


src/modules/visitor-space/components/AutocompleteFieldFilterForm/AutocompleteFieldFilterForm.tsx
```diff
@@ -0,0 +1,105 @@
...
+					setForm({ value: '' });
+					clearErrors();
+				},
+				handleSubmit,
+			})}
+		</>
+	);
+};
+
+export default AutocompleteFieldFilterForm;
```
RobbeBierebeeck, line 105: I prefer to export the component with a named export directly from the place where it is defined

src/modules/visitor-space/components/AutocompleteFieldFilterForm/AutocompleteFieldFilterForm.tsx
```diff
@@ -0,0 +1,105 @@
...
+import { FormControl } from '@meemoo/react-components';
+import clsx from 'clsx';
+import { isNil } from 'lodash-es';
+import { type FC, type ReactNode, useEffect, useState } from 'react';
+import { Controller, useForm } from 'react-hook-form';
+import { StringParam, useQueryParam } from 'use-query-params';
+import { object as yupObject, string as yupString } from 'yup';
+
+import AutocompleteFieldInput from '@visitor-space/components/AutocompleteFieldInput/AutocompleteFieldInput';
+import { AutocompleteField } from '@visitor-space/components/FilterMenu/FilterMenu.types';
```
RobbeBierebeeck, line 11: I would shorten the import path a bit, so these two components can come from the same import path. 
`import { AutocompleteFieldInput, AutocompleteField } from '@visitor-space/components'`

src/modules/visitor-space/components/AutocompleteFieldInput/AutocompleteFieldInput.module.scss
```diff
@@ -0,0 +1,12 @@
+@use "src/styles/abstracts" as *;
+
+$component: "c-autocomplete-field-input";
+
+.#{$component} {
+	width: 100%;
```
RobbeBierebeeck, line 6: I'm not a fan of defining this class in a different component variable. It just makes it more complicated for no reason.

src/modules/visitor-space/components/AutocompleteFieldInput/AutocompleteFieldInput.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+const AutocompleteFieldInput: FC<AutocompleteFieldInputProps> = ({
+	onChange,
+	value,
+	fieldName,
+	label,
+}) => {
+	const handleLoadOptions = useCallback(
+		(inputValue: string, callback: (newOptions: SelectOption[]) => void): void => {
+			if (inputValue.length < 3) {
```
RobbeBierebeeck, line 30: Is this 3 we can maybe define in a different const?

src/modules/visitor-space/components/AutocompleteFieldInput/AutocompleteFieldInput.tsx
```diff
@@ -0,0 +1,77 @@
...
+	);
+
+	const handleChange = (
+		newValue: SingleValue<SelectOption>,
+		actionMeta: ActionMeta<SelectOption>
+	): void => {
+		if (actionMeta.action === 'select-option' || actionMeta.action === 'create-option') {
+			onChange(newValue?.value || null);
+		}
+	};
```
RobbeBierebeeck, line 48: Is there a need for this different handleChange component. Can't this be don direcly from the `<AsyncCreatableSelect<SelectOption>` component

src/modules/visitor-space/components/DurationFilterForm/DurationFilterForm.tsx
```diff
@@ -114,7 +114,8 @@ const DurationFilterForm: FC<DurationFilterFormProps> = ({ children, className,
 
 											if (value !== form.operator) {
 												setForm({
-													duration: defaultValues.duration,
+													[SearchFilterId.Duration]:
+														defaultValues.duration,
```
RobbeBierebeeck, line 118: It feels a bit inconsistent that the first duration starts with a capital letter and the second duration not

RobbeBierebeeck (approved):

src/modules/visitor-space/components/AutocompleteFieldFilterForm/AutocompleteFieldFilterForm.tsx
bertyhell: that requires a barrel file
and those often have high performance impact on tests
https://marvinh.dev/blog/speeding-up-javascript-ecosystem-part-7/
https://dev.to/tassiofront/barrel-files-and-why-you-should-stop-using-them-now-bc4

src/modules/visitor-space/components/AutocompleteFieldInput/AutocompleteFieldInput.module.scss
bertyhell: i agree
but it is the current coding style of the repo
if we decide we want to change it we need to agree with all devs that work on the project before we start changing the code

src/modules/visitor-space/components/AutocompleteFieldInput/AutocompleteFieldInput.tsx
bertyhell: you mean: inline in the html code?
i think its cleaner to keep the html code as minimal as possible and make any onChange that isn't just calling the onChange of the props a separate handler in the code.

src/modules/visitor-space/components/DurationFilterForm/DurationFilterForm.tsx
bertyhell: one is an enum
the second is a value

There isn't a clear enum casing convention in the code base

Which do you prefer?
I kind of like all caps to signify its a constant

SearchFilterId.DURATION

PR #1269, opened 2024-10-02
fix(ARC-2496): change metadata order and rendering + split object detail

bertyhell:
https://meemoo.atlassian.net/browse/ARC-2496

split detail object page in 2:
* ObjectDetailPage (1600 lines)
* ObjectDetailPageMetadata (1000 lines)

also handled some errors resulting from the latest admin-core version

image


bertyhell:

PR #1271, opened 2024-10-03
fix(ARC-2415): fix add to folder checkboxes, remove some useForm hooks

bertyhell:
https://meemoo.atlassian.net/browse/ARC-2415

* fix add to folder checkbox issue where you can't uncheck already checked folder
* remove some useForm hooks
    * since the useForm library throws some strange errors:
        * `Unexpected return value from a callback ref in input. A callback ref should not return a function.`
    * and the library is quite verbose for minimal gains
        * with or without this library, we use the same amount of code lines
* Add manual form validation function using joi schemas

image

### DIFF
--- a/src/modules/account/views/MyProfile.tsx
+++ b/src/modules/account/views/MyProfile.tsx
@@ -6,7 +6,6 @@ import {
 	Button,
 	Checkbox,
 	CheckboxList,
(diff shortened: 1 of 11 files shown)


bertyhell:

PR #1272, opened 2024-10-04
fix(styles): fix some minor style issues/improvements

bertyhell:
Add dropshadow behind dropdown menu:
image

Fix export button label on desktop/mobile
image

Fix z-index for iiif viewer showing up above the backdrop of modals
image


bertyhell:

PR #1279, opened 2025-01-22
fix(ARC-1920): add check for manage own folder permissions

RobbeBierebeeck:


src/modules/navigation/components/Navigation/Navigation.consts.tsx
```diff
@@ -542,8 +544,15 @@ export const getNavigationItemsProfileDropdown = (
 		}
 	);
 
+	const filteredDefaultRoutes = defaultRoutes?.filter((navItem: NavigationItem) => {
+		return !(
+			navItem.path.trim() === ROUTES_BY_LOCALE[locale].accountMyFolders.trim() &&
+			!user?.permissions?.includes(Permission.MANAGE_FOLDERS)
+		);
```
bertyhell, line 551: this is a bit too much logic in one line with double negation
```suggestion
		if (navItem.path.trim() === ROUTES_BY_LOCALE[locale].accountMyFolders.trim()) {
		    // Check managed folders permission
		    return user?.permissions?.includes(Permission.MANAGE_FOLDERS);
		}
		// All other nav items can be shown
		return true;
```

bertyhell (approved):

bertyhell:
please add a link and a screenshot to the PR

PR #1284, opened 2025-03-03
feat(ARC-2681): unify filter form interfaces

bertyhell:


bertyhell:
This change is too large and causes too many issues to merge
We're closing it for now 😢

PR #1421, opened 2026-04-07
fix: Avoid the detail blade keeping the scroll lock active on closing…

reunefe:
… because it does not render the blademanager without an item

Steps to reproduce:
- Go to my material requests
- open a detail blade
- close the detail blade

=> scroll lock is still active but it should not

### DIFF
--- a/src/modules/account/components/MaterialRequestDetailBlade/MaterialRequestDetailBlade.tsx
+++ b/src/modules/account/components/MaterialRequestDetailBlade/MaterialRequestDetailBlade.tsx
@@ -440,24 +440,24 @@ const MaterialRequestDetailBlade: FC<MaterialRequestDetailBladeProps> = ({
 					'modules/account/components/material-request-detail-blade/material-requests___detail'
 				),
 				stickySubtitle: <MaterialRequestInformation />,
-				subtitle: (
+				subtitle: currentMaterialRequestDetail ? (
 					<MaterialCard
 						openInNewTab={true}
-						objectId={currentMaterialRequestDetail?.objectSchemaIdentifier}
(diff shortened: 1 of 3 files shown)


src/modules/account/components/MaterialRequestDetailBlade/MaterialRequestDetailBlade.tsx
```diff
@@ -436,17 +443,17 @@ const MaterialRequestDetailBlade: FC<MaterialRequestDetailBladeProps> = ({
 				subtitle: (
 					<MaterialCard
 						openInNewTab={true}
-						objectId={currentMaterialRequestDetail.objectSchemaIdentifier}
-						title={currentMaterialRequestDetail.objectSchemaName}
-						thumbnail={currentMaterialRequestDetail.objectThumbnailUrl}
+						objectId={currentMaterialRequestDetail?.objectSchemaIdentifier}
+						title={currentMaterialRequestDetail?.objectSchemaName}
+						thumbnail={currentMaterialRequestDetail?.objectThumbnailUrl}
```
bertyhell: isn't it safer to do instead of rendering with partial data?
```
				subtitle: (
					currentMaterialRequestDetail ? <MaterialCard ...> : null
                )
```

bertyhell (approved):

src/modules/account/components/MaterialRequestDetailBlade/MaterialRequestDetailBlade.tsx
reunefe: True, ik had het gewoon snel opgelost puur omdat dit nooit gaat voorkomen 😆

PR #1440, opened 2026-04-20
fix(ARC-3531): disable rich text editor buttons when rte is disabled

bertyhell:
https://meemoo.atlassian.net/browse/ARC-3531

before:
image

after:
image

### DIFF
--- a/src/modules/account/components/MaterialRequestDetailBlade/MaterialRequestConversation.module.scss
+++ b/src/modules/account/components/MaterialRequestDetailBlade/MaterialRequestConversation.module.scss
@@ -199,7 +199,8 @@ $rich-text-editor-height-mobile: 14.7rem; // rich text editor
 
 		&--disabled {
 			:global(.control-item.button),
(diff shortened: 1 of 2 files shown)


reunefe (approved):

bertyhell:
@reunefe the focus doesn't seem to loop anymore. so i wonder what is worse:
* looping through the rte buttons and getting back to the top of the page
* or not looping through the buttons, but having to use shift + tab to get back to the top

i would say option 1 is better

reunefe:
> @reunefe the focus doesn't seem to loop anymore. so i wonder what is worse:
> 
> * looping through the rte buttons and getting back to the top of the page
> * or not looping through the buttons, but having to use shift + tab to get back to the top
> 
> i would say option 1 is better

The problem of the not looping was actually because the buttons still got an invisible focus. I removed this now by setting the tabIndex to -1

Senn-Patteet (approved):

PR #1444, opened 2026-04-20
fix(ARC-3503): refetch material request when download becomes available

bertyhell:
https://meemoo.atlassian.net/browse/ARC-3503

before: materialRequest.downloadStatus can get out of sync with events loaded in the conversation:
image

after:
TODO test this PR


reunefe (approved):
Dit werkt, nice catch!

PR #1469, opened 2026-04-23
Feature/arc 3630 add archivation date to final message

reunefe:


src/modules/account/components/MaterialRequestDetailBlade/MaterialRequestConversationMessage.tsx
```diff
@@ -140,6 +145,23 @@ export const MaterialRequestConversationMessage: FC<MaterialRequestConversationM
 		);
 	};
 
+	const renderFinalSummary = () => {
+		if (!materialRequest.archivationDate) {
```
bertyhell: we have:
* createdAt
* updatedAt
* requestedAt

i think we should rename this one to:
* archivedAt

bertyhell (approved):

src/modules/account/components/MaterialRequestDetailBlade/MaterialRequestConversationMessage.tsx
reunefe: I disagree, the date here will be used for something in the future.
Although the consistency would be nice, the date itself is just to reference when something will happen, not that it happened

reunefe: Maybe `willBeArchivedAt`? It's just something build on the proxy with an env variable. Nothing saved in the database

bertyhell: why do we store this?
shouldn't it just be:
archived_at: null => when not yet archived
archived_at: date => when archived at that date

the daily check for archived material requests, can just check the final event date and set the archived_at date when it is being archived

otherwise we won't be able to change the MATERIAL_REQUEST_TIME_BEFORE_ARCHIVATION since that value will be baked into every material request

bertyhell (changes requested):
see comment

src/modules/account/components/MaterialRequestDetailBlade/MaterialRequestConversationMessage.tsx
reunefe: We do not store this. That's the thing, we do not need to. The archivationDate is a computed value based on the created_at of the final summary event. It takes that value and adds he MATERIAL_REQUEST_TIME_BEFORE_ARCHIVATION value. We shouldn't store it.

However we need that computed field in the client for display reasons so the user knows when it will be archived. That's the only reason this property is added. That's why I don't like the naming to change to archived_at and would like to keep it as is

bertyhell (approved):

PR #1472, opened 2026-04-24
feat(archived): show archived indicator for material requests

bertyhell:
it seems a bit strange to not show any indicator for material requests that havee been archived
since checking the filter or unchecking it, just shows some more or less material requests, but you don't see which is which

I suspect they will want to show an indicator
here is the PR for that. But let's not merge it yet
image


reunefe:
I really disagree with the indicator. Because it will be on every request in the list, or on none because of the toggle 😂

Maybe just change the title on the page or something ?

bertyhell:
> I really disagree with the indicator. Because it will be on every request in the list, or on none because of the toggle 😂
> 
> Maybe just change the title on the page or something ?

ah, true
ok, let's leave it as is. And see what they say

PR #1476, opened 2026-04-28
fix: remove unused preconnect tags

Senn-Patteet:


src/pages/_document.tsx
```diff
@@ -67,8 +67,6 @@ j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src=
 				<link rel="preconnect" href="https://ekr.zdassets.com" />
 				<link rel="preconnect" href="https://www.googletagmanager.com" />
 				<link rel="preconnect" href="https://consent.cookiebot.com" />
-				<link rel="preconnect" href="https://consentcdn.cookiebot.com" />
```
bertyhell, line 70: i still see this one on qas
so i think we should leave it
you don't see it in the network tab?

Image

bertyhell (changes requested):

bertyhell (approved):

PR #1477, opened 2026-04-28
fix: padding + search icon

Senn-Patteet:


bertyhell (approved):

bertyhell:
does this also look good on mobile devices?

reunefe (approved):

PR #1488, opened 2026-04-30
tfeat: add herbruikbaarheid filter

RobbeBierebeeck:


src/modules/visitor-space/components/ReusabilityFilterForm/ReusabilityFilterForm.const.ts
```diff
@@ -9,17 +9,17 @@ export const REUSABILITY_FILTER_FORM_QUERY_PARAM_CONFIG = {
 
 export const REUSABILITY_OPTIONS = () => [
```
bertyhell: i usually use GET_ for functions, so the developer knows it's not a constant variable, but a function that returns constants just by looking at the name
```suggestion
export const GET_REUSABILITY_OPTIONS = () => [
```

src/modules/visitor-space/components/ReusabilityFilterForm/ReusabilityFilterForm.const.ts
```diff
@@ -9,17 +9,17 @@ export const REUSABILITY_FILTER_FORM_QUERY_PARAM_CONFIG = {
...
 	{
-		key: 'herbruikbaar-onder-voorwaarden',
+		key: 'reusable-with-conditions',
 		label: tText(
 			'modules/visitor-space/components/reusability-filter-form___herbruikbaar-onder-voorwaarden'
 		),
 	},
 	{
-		key: 'misschien-herbruikbaar',
+		key: 'possibly-reusable',
```
bertyhell: might be nice to have these as a enum, that we can also copy to the client, so they stay in sync.
At some point, if we ever make this repo a mono repo, it's then easier to move the client/proxy enums and types to a shared folder

bertyhell (approved):

src/modules/shared/types/ie-objects.ts
```diff
@@ -48,7 +48,7 @@ export enum IeObjectsSearchFilterField {
 	CONSULTABLE_ONLY_ON_LOCATION = 'isConsultableOnlyOnLocation',
 	CONSULTABLE_MEDIA = 'isConsultableMedia',
 	CONSULTABLE_PUBLIC_DOMAIN = 'isConsultablePublicDomain',
-	REUSABILITY = 'reusability',
```
bertyhell: isn't this a revert from a previous fix to make it english?

src/modules/visitor-space/const/visitor-space-filters.const.tsx
```diff
@@ -130,11 +127,6 @@ export const SEARCH_PAGE_FILTERS = (
 		form: ReusabilityFilterForm,
 		type: FilterMenuType.Modal,
 		tabs: ALL_TABS,
-		isDisabled: () => {
-			const rightsFiltersForEverybody =
-				publicRuntimeConfig.ENABLE_RIGHTS_FILTERS_FOR_EVERYBODY === 'true';
-			return !rightsFiltersForEverybody && !isKeyUser;
-		},
```
bertyhell: don't delete this env var, since we might need it for gradual rollout to production

bertyhell (changes requested):

bertyhell (approved):

PR #1498, opened 2026-05-07
feat(ie-objects): add rights label display and restrict ticket reques

RobbeBierebeeck:
…ts by object type (ARC-3594)


src/modules/ie-objects/components/ObjectDetailPageMetadata/ObjectDetailPageMetadata.tsx
```diff
@@ -1000,12 +1000,12 @@ export const ObjectDetailPageMetadata: FC<ObjectDetailPageMetadataProps> = ({
 							</span>
 						</Metadata>
 					)}
-					{!!avRightsInfo?.reuseLabel && (
+					{avRightsInfo !== undefined && (
```
bertyhell: unless you're checking a number, you can just cast the object to boolean for a check so it also handles null
```suggestion
					{!!avRightsInfo && (
```

bertyhell (approved):

PR #1509, opened 2026-05-11
Feature/optimizations

reunefe:
Verwijderen van ongebruikte code, image optimalisaties


Senn-Patteet (approved):

src/modules/ie-objects/components/RelatedObject/RelatedObject.tsx
```diff
@@ -19,7 +19,6 @@ const RelatedObject: FC<RelatedObjectProps> = ({ className, object }) => {
 				return (
 					<CardImage
 						className={styles['c-related-object__image']}
-						unoptimized
```
bertyhell, line 22: i thought that these were also trying to optimize images that live on other servers, which isn't possible.
only local image urls that nextjs hosts can be optimized i think, so you might want to double check before deleting these everywhere

src/modules/ie-objects/services/ie-objects/ie-objects.service.ts
```diff
@@ -236,14 +236,6 @@ export class IeObjectsService {
 			.json();
 	}
 
-	public static async lookupNvdgoId(
-		mediaMosaId: string
-	): Promise<{ schema_identifier: string; title: string; maintainerSlug: string }> {
-		return await ApiService.getApi()
-			.get(`${IE_OBJECTS_SERVICE_BASE_URL}/lookup/nvdgo/${mediaMosaId}`)
-			.json();
-	}
```
bertyhell, line 245: this sounds like something we need to provide redirect for old "Nieuws van de groote oorlog" website
if it is no longer used, something might be wrong there?

it should be related to these redirects:
https://github.com/viaacode/hetarchief-client/blob/1c93222de60e5bd995f12ca3cca017c04ca4c6c2/next.config.js#L398

src/modules/iiif-viewer/IiifViewer.tsx
```diff
@@ -943,15 +944,15 @@ export const IiifViewer = ({
 								className={activeImageIndex === index ? 'active' : ''}
 							>
 								{imageInfo.thumbnailUrl ? (
-									// biome-ignore lint/performance/noImgElement: this is how it works
-									<img
+									<Image
```
bertyhell, line 947: are we sure these work for external images? (images not hosted by the nextjs service)

bertyhell (approved):

bertyhell (changes requested):
nvdgo id's urls should still be redirected, if that function is unued, something is broken and should be fixed

src/modules/ie-objects/components/RelatedObject/RelatedObject.tsx
reunefe: I actually removed this and set the option globally

src/modules/ie-objects/services/ie-objects/ie-objects.service.ts
reunefe: The method is no where used. I globally searched on lookupNvdgoId and the only hit is this method 🤔

src/modules/iiif-viewer/IiifViewer.tsx
reunefe: Everything seems to be still working, but I'll double check

src/modules/ie-objects/services/ie-objects/ie-objects.service.ts
bertyhell: hmm, probably i forgot
normally it should be a page in nextjs under /pages/media/[nvdgoId]
and inside it should do the
```typescript
lookupNvdgoId(nvdgoId) => {
	schema_identifier,
	title,
	maintainerSlug,
}
```
which then redirects to /zoeken/maintainer-slug/schema_identifer/kebabCase(title)

reunefe: There is no page for media. However we have in the middleware the following:
```
if (
	pathName.startsWith('/media/') ||
	pathName.startsWith('/nl/media/') ||
	pathName.startsWith('/en/media/')
) {
	const pathParts = pathName.split('/media/');
	const mediaMosaId = pathParts.pop()?.split('/').pop();
	if (!mediaMosaId) {
		console.error('No mediaMosaId found in the URL: ', pathName);
		return NextResponse.next();
	}
	const ieObjectRedirectInfo: IeObjectRedirectInfo = (await (
		await fetch(`${process.env.PROXY_URL}/ie-objects/lookup/nvdgo/${mediaMosaId}`)
	).json()) as IeObjectRedirectInfo;
	const schemaIdentifier = ieObjectRedirectInfo.schema_identifier;
	if (!schemaIdentifier) {
		console.error(
			'Fetching the schemaIdentifier for the nvdgo id failed: ',
			ieObjectRedirectInfo
		);
		return NextResponse.rewrite(new URL('/__force-404', request.url));
	}
	const search = pathName.startsWith(`/${Locale.En}/`)
		? `${Locale.En}/${ROUTE_PARTS_BY_LOCALE.en.search}`
		: ROUTE_PARTS_BY_LOCALE.nl.search;
	const maintainerSlug = ieObjectRedirectInfo.maintainerSlug;
	const title = ieObjectRedirectInfo.title ? ieObjectRedirectInfo.title.replace(/ /g, '-') : '';

	return NextResponse.redirect(
		new URL(`/${search}/${maintainerSlug}/${schemaIdentifier}/${title}`, request.url)
	);
}
```

bertyhell (approved):

PR #1515, opened 2026-05-13
Feature/arc 3598  make reusability filter available for all users

RobbeBierebeeck:


src/modules/visitor-space/utils/elastic-filters/elastic-filters.test.ts
```diff
@@ -0,0 +1,24 @@
...
+				`vrij-herbruikbaar${FILTER_LABEL_VALUE_DELIMITER}Vrij herbruikbaar`,
+				`herbruikbaar-onder-voorwaarden${FILTER_LABEL_VALUE_DELIMITER}Herbruikbaar onder voorwaarden`,
+				`misschien-herbruikbaar${FILTER_LABEL_VALUE_DELIMITER}Misschien herbruikbaar`,
+			],
+		} as SearchPageQueryParams);
+
+		expect(filters.find(({ field }) => field === IeObjectsSearchFilterField.REUSABILITY)).toEqual({
+			field: IeObjectsSearchFilterField.REUSABILITY,
+			operator: IeObjectsSearchOperator.IS,
+			multiValue: ['vrij-herbruikbaar', 'herbruikbaar-onder-voorwaarden', 'misschien-herbruikbaar'],
```
bertyhell: might be nicer if this is en enum that also exists in the proxy

next.config.js
```diff
@@ -128,7 +128,6 @@ module.exports = withTM({
 		ENABLE_GOOGLE_INDEXING: process.env.ENABLE_GOOGLE_INDEXING,
 		IIIF_IMAGE_API: process.env.IIIF_IMAGE_API,
 		ENABLE_MATERIAL_REQUEST_COMPLEX_REUSE_FLOW: process.env.ENABLE_MATERIAL_REQUEST_COMPLEX_REUSE_FLOW,
-		ENABLE_RIGHTS_FILTERS_FOR_EVERYBODY: process.env.ENABLE_RIGHTS_FILTERS_FOR_EVERYBODY,
```
bertyhell, line 131: why is this deleted?

src/modules/visitor-space/const/visitor-space-filters.const.tsx
```diff
@@ -111,18 +111,22 @@ export const SEARCH_PAGE_FILTERS = (
...
 		label: tText('modules/visitor-space/const/visitor-space-filters___herbruikbaarheid'),
 		form: ReusabilityFilterForm,
 		type: FilterMenuType.Modal,
 		tabs: ALL_TABS,
-		isDisabled: () => {
-			const { publicRuntimeConfig } = getConfig();
-			const rightsFiltersForEverybody =
-				publicRuntimeConfig.ENABLE_RIGHTS_FILTERS_FOR_EVERYBODY === 'true';
-			return !rightsFiltersForEverybody && !isKeyUser;
-		},
```
bertyhell, line 137: why is this deleted?

bertyhell (changes requested):

next.config.js
RobbeBierebeeck: We removed ENABLE_RIGHTS_FILTERS_FOR_EVERYBODY because ARC-3598 makes the reusability filter available to all users unconditionally. The env flag was only used to decide whether the filter should be hidden for non-key users; after removing that visibility check, the config value was no longer read anywhere, so keeping it would be dead configuration.

src/modules/visitor-space/const/visitor-space-filters.const.tsx
RobbeBierebeeck: This was removed because ARC-3598 changes the filter from feature-flagged/key-user-only to always visible for all users. This isDisabled block was the only place where ENABLE_RIGHTS_FILTERS_FOR_EVERYBODY was used; with the new requirement, the reusability filter should no longer be disabled for non-key users, so the condition became obsolete.

next.config.js
bertyhell: meemoo had gevraagd om die env var toe te voegen omdat ze in het begin nog geen rechten info gaan hebben voor objecten die alle gebruikers mogen zien. Dus graag terug toevoegen

PR #1527, opened 2026-05-20
feat(ARC-3595): show AV rights section on object detail

RobbeBierebeeck:


src/modules/ie-objects/components/ObjectDetailPageMetadata/ObjectDetailPageMetadata.tsx
```diff
@@ -843,6 +848,9 @@ export const ObjectDetailPageMetadata: FC<ObjectDetailPageMetadataProps> = ({
 		const avRightsInfo = AV_OBJECT_TYPES.includes(mediaInfo.dctermsFormat)
 			? mediaInfo.rightsInfo
 			: null;
+		const avRightsIcon = avRightsInfo ? getIeObjectAvRightsIcon(avRightsInfo) : null;
+		const avRightsLabel = avRightsInfo ? getIeObjectAvRightsLabel(avRightsInfo) : null;
+		const avRightsUrl = avRightsInfo ? getIeObjectAvRightsUrl(avRightsInfo) : undefined;
```
bertyhell: can't the helper functions handle the undefined avRightsInfo and return null instead of doing this outside of the function?

src/modules/ie-objects/utils/get-ie-object-av-rights-icon.tsx
```diff
@@ -0,0 +1,111 @@
...
+
+	if (rightsText.includes('cc0') || rightsText.includes('creativecommons.org/publicdomain')) {
+		return tText('modules/visitor-space/const/rights-filter___cc0');
+	}
+
+	if (rightsText.includes('/cne/') || rightsText.includes('bescherming niet bepaald')) {
+		return tText(
+			'modules/visitor-space/const/rights-filter___auteursrechtelijke-bescherming-niet-bepaald'
+		);
+	}
```
bertyhell, line 108: this seems strange
why not do this with a lookup table for each rights id url
or are those not available?

or is this logic described in the FA?

src/modules/ie-objects/utils/get-ie-object-av-rights-icon.tsx
```diff
@@ -0,0 +1,111 @@
...
+		return tText('modules/visitor-space/const/rights-filter___auteursrechtelijk-beschermd');
+	}
+
+	return rightsInfo.reuseCategoryLabel || rightsInfo.reuseLabel;
+}
+
+export function getIeObjectAvRightsIcon(rightsInfo: IeObjectRightsInfo): ReactNode | null {
+	const rightsText = getRightsSearchText(rightsInfo);
+
+	if (!rightsText || rightsText.includes('geen rechteninformatie')) {
```
bertyhell, line 135: might be safer?
```suggestion
	if (!rightsText || rightsText.toLowerCase().includes('geen rechteninformatie')) {
```

bertyhell (approved):

PR #1528, opened 2026-05-20
Feature/arc 3544 redirect all slugs to new table

reunefe:


src/modules/cp/components/VisitorSpaceSettings/VisitorSpaceSettings.tsx
```diff
@@ -285,12 +285,6 @@ const VisitorSpaceSettings: FC<VisitorSpaceSettingsProps> = ({ action, visitorSp
 					});
 
 					await refetchVisitorSpace();
-					if (!!values.slug && values.slug !== visitorSpace?.slug) {
-						// Slug was changed, redirect to the new url
-						await router.replace(
-							`/${ROUTE_PARTS_BY_LOCALE[locale].admin}/${ROUTE_PARTS_BY_LOCALE[locale].visitorSpaceManagement}/${ROUTE_PARTS_BY_LOCALE[locale].visitorSpaces}/${values.slug}`
-						);
-					}
```
bertyhell, line 293: don't we need this anymore? if the slug changes redirect the user to the new visitor space url

bertyhell (approved):

src/modules/cp/components/VisitorSpaceSettings/VisitorSpaceSettings.tsx
reunefe: No, because the slug is not editable here anymore. That is moved to those new admin screens

PR #1538, opened 2026-05-26
feat(ARC-3606): show AV bronvermelding

RobbeBierebeeck:


src/modules/ie-objects/components/ObjectDetailPageMetadata/ObjectDetailPageMetadata.tsx
```diff
@@ -839,131 +839,51 @@ export const ObjectDetailPageMetadata: FC<ObjectDetailPageMetadataProps> = ({
...
-			? getIeObjectRightsStatusInfo(mediaInfo, locale)
-			: null;
-
+		const shouldShowAvRightsAttribution =
+			AV_OBJECT_TYPES.includes(mediaInfo.dctermsFormat) &&
+			[
+				IeObjectLicense.PUBLIEK_CONTENT,
+				IeObjectLicense.INTRA_CP_CONTENT,
+				IeObjectLicense.BEZOEKERTOOL_CONTENT,
+			].some((license) => mediaInfo?.licenses?.includes(license));
```
bertyhell: we usually check if the user has acceess to the video/audio
by checking if there is a thumbnail url on the object
since the proxy strips that url if the user doesn't have access

checking the licenses is not a garantee that they have access to the video/audio

src/modules/ie-objects/utils/get-ie-object-av-rights-attribution-text.ts
```diff
@@ -0,0 +1,132 @@
...
+
+function getLocalizedMissingRightsInfo(locale: Locale): string {
+	return locale === Locale.en
+		? 'no rights information available'
+		: 'geen rechteninformatie beschikbaar';
+}
+
+function getLocalizedAnd(locale: Locale): string {
+	return locale === Locale.en ? 'and' : 'en';
+}
```
bertyhell: use the translation functions:
```
tText('Unknown creator')
tText('and')
```

those auto translate using the active locale
and those are included in the extraction script for extracting traanslations, so  meemoo can manage the translations themselves

src/modules/ie-objects/utils/get-ie-object-av-rights-attribution-text.ts
```diff
@@ -0,0 +1,132 @@
...
+
+	if (uniqueNames.length === 2) {
+		return `${uniqueNames[0]} ${getLocalizedAnd(locale)} ${uniqueNames[1]}`;
+	}
+
+	if (uniqueNames.length === 3) {
+		return `${uniqueNames[0]}, ${uniqueNames[1]} ${getLocalizedAnd(locale)} ${uniqueNames[2]}`;
+	}
+
+	return `${uniqueNames.slice(0, 3).join(', ')}, e.a.`;
```
bertyhell: use tText to translate the "e.a."

bertyhell (approved):

bertyhell (approved):

PR #1540, opened 2026-05-27
feat: add source attribution and provider purl metadata

RobbeBierebeeck:


src/modules/ie-objects/components/ObjectDetailPageMetadata/ObjectDetailPageMetadata.module.scss
```diff
@@ -73,6 +73,39 @@ $maintainer-logo-height: 7rem;
...
+}
+
+.p-object-detail__provider-identifier-link {
+	color: inherit;
+	text-decoration: underline;
+	text-underline-offset: 0.2rem;
+
+	&:hover,
+	&:focus {
+		color: inherit;
```
bertyhell: isn't that the default?

src/modules/ie-objects/components/ObjectDetailPageMetadata/ObjectDetailPageMetadata.module.scss
```diff
@@ -73,6 +73,39 @@ $maintainer-logo-height: 7rem;
...
+	}
+
+	&:focus-visible {
+		outline: 0.2rem solid $teal;
+		outline-offset: 0.2rem;
+	}
+}
+
+.p-object-detail__provider-identifier-link {
+	color: inherit;
```
bertyhell: isn't that the default?

bertyhell (approved):

PR #1553, opened 2026-06-02
Bugfix/arc 3712 additional conditions

reunefe:


src/modules/account/components/MaterialRequestDetailBlade/MaterialRequestContentInfo.tsx
```diff
@@ -142,7 +142,6 @@ const MaterialRequestContentInfo: FC<MaterialRequestContentInfoProps> = ({
 						}
 					);
 				case MaterialRequestEventType.CANCELLED:
-				case MaterialRequestEventType.ADDITIONAL_CONDITIONS_DENIED:
```
bertyhell, line 145: maybe this case should be commented with an additional comment: 
We don't render this message, since the denied event is enough to be displayed in the chat. We don't need 2 messages for this

src/modules/account/components/MaterialRequestDetailBlade/MaterialRequestConversationMessage.tsx
```diff
@@ -43,6 +44,25 @@ export const MaterialRequestConversationMessage: FC<MaterialRequestConversationM
 }) => {
 	const user = useSelector(selectCommonUser);
 
+	// In case the message is the approved event and there were additional conditions with auto-approval, no need to render this message
+	if (message.messageType === MaterialRequestEventType.APPROVED) {
+		const hasConditions = materialRequest.history.find(
+			(event) => event.messageType === MaterialRequestEventType.ADDITIONAL_CONDITIONS
```
bertyhell, line 50: isn't it more logical to get autoApproveAfterAcceptAdditionalConditions from the ADDITIONAL_CONDITIONS and also check that there is a ADDITIONAL_CONDITIONS_ACCEPTED event?

implicilty the existinace of ADDITIONAL_CONDITIONS  and APPROVED means the event ADDITIONAL_CONDITIONS_ACCEPTED also exists. But an explicit check seems more clear to me

bertyhell (approved):

src/modules/account/components/MaterialRequestDetailBlade/MaterialRequestConversationMessage.tsx
reunefe: Niet echt.
We willen de message van APPROVE niet renderen als de auto-approve gezet is. Ongeacht hoe die approval er gekomen is. Anders zou je mogelijks die boodschap wel tonen, ondanks auto-approve

PR #1564, opened 2026-06-09
Bugfix/arc 3394 infinite loading

reunefe:


src/modules/search/MaintainerSearchPage.tsx
```diff
@@ -60,5 +61,9 @@ export const MaintainerSearchPage: FC<MaintainerSearchPageProps> = () => {
 		}
 	}, [router, ieObjectInfo, locale]);
 
+	if (isErrorOrganisation && isErrorObject) {
+		return <ErrorNotFound />;
```
bertyhell, line 65: if we want this to be a hard 404, shouldn't we also fetch these requests in the server side rendering: https://github.com/viaacode/hetarchief-client/blob/bb5f30dfc888216d1c7747bc8a0a4aeff1a5a265/src/pages/zoeken/%5Bslug%5D/index.tsx#L15

And return a 404 there?

similar to this detail page: https://github.com/viaacode/hetarchief-client/blob/38e177d0e2def4b2261cb349f4f6e238f4649449/src/pages/zoeken/%5Bslug%5D/%5Bie%5D/%5Bname%5D/index.tsx#L36

bertyhell (approved):
We can already merge this
the SSR 404 we could do in a separate PR

src/modules/search/MaintainerSearchPage.tsx
reunefe: Yeah, technically everything should move indeed to the SSR, I'll change it although that was a bigger change than the quick fix 😅

reunefe: [See PR](https://github.com/viaacode/hetarchief-client/pull/1566)

PR #1565, opened 2026-06-10
ARC-3438: set languge checkboxes correctly

reunefe:


src/modules/account/views/AccountMyProfile.tsx
```diff
@@ -87,10 +90,10 @@ export const AccountMyProfile: FC<DefaultSeoInfo> = ({ url, canonicalUrl }) => {
 	const canEdit = Object.keys(commonUser?.idps || {}).includes(Idp.HETARCHIEF) && canEditProfile;
 
 	useEffect(() => {
-		if (currentAccountLocale && !selectedLanguage) {
+		if (currentAccountLocale) {
 			setSelectedLanguage(currentAccountLocale);
 		}
-	}, [currentAccountLocale, selectedLanguage]);
+	}, [currentAccountLocale]);
```
bertyhell, line 96: isn't this the same as this without the need for the useMemo?
```
	useEffect(() => {
		setSelectedLanguage((commonUser?.language || Locale.nl) as Locale);
	}, [commonUser?.language]);
```

bertyhell (approved):

src/modules/account/views/AccountMyProfile.tsx
reunefe: True, but the currentAccountLocale is used in a multitude of places, so I do not see why we would change the effect 🤔

PR #1590, opened 2026-06-29
ARC-3747: remove user data from events since this will be done on de …

reunefe:
…proxy


src/modules/ie-objects/ObjectDetailPage.tsx
```diff
@@ -958,20 +956,24 @@ export const ObjectDetailPage: FC<DefaultSeoInfo> = ({
 	/**
 	 * Trigger events for viewing the ie object
 	 */
+
+	const mapObjectToEventData = useCallback(() => {
```
bertyhell:
```suggestion
	const mapIeObjectToEventData = useCallback(() => {
```

bertyhell (approved):

PR #1598, opened 2026-07-10
Feature/lcm upgrade

reunefe:


.storybook/preview.tsx
```diff
@@ -0,0 +1,50 @@
...
+import type { Preview } from '@storybook/nextjs';
+import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
+import React from 'react';
+import { Provider } from 'react-redux';
+
+import NextQueryParamProvider from '../src/modules/shared/providers/NextQueryParamProvider/NextQueryParamProvider';
+import { makeStore } from '../src/modules/shared/store/store';
+
+// next/image is handled automatically by the @storybook/nextjs framework, so the previous
+// manual unoptimized-image override is no longer needed.
```
bertyhell: can be deleted i guess

src/styles/abstracts/_functions.scss
```diff
@@ -26,3 +26,36 @@
...
+	@return '[class*="#{$component}_#{$local-class}#{$suffix}"]';
+}
+
+// Turbopack separates the component name from the local class with an unpredictable hash
+// (`Component-module-scss-module__hash__local-class`), so instead of one contiguous string we
+// chain two attribute selectors -- each anchored to a fixed, hash-independent part -- which must
+// both match the same class attribute regardless of what sits between them.
+@function module-class-selector-turbopack($component, $local-class, $suffix: "__") {
+	@return '[class*="#{$component}-module-scss-module#{$suffix}"][class*="#{$suffix}#{$local-class}"]';
+}
```
bertyhell, line 54: if we got turbopack working, do we still need the webpack function?

bertyhell (approved):

src/styles/abstracts/_functions.scss
reunefe: Probably not, but I left in here just in case we would need to revert. Luckily it would be an easy removal

PR #1600, opened 2026-07-16
feat (Carousel): Adjust config with new block

reunefe:


reunefe:
Desktop:
https://github.com/user-attachments/assets/387974da-e1bf-4542-b44a-d145c9d3dd76

Mobiel:
https://github.com/user-attachments/assets/7504ff35-65be-4ac1-b55b-428735c090e4

package.json
```diff
@@ -8,7 +8,8 @@
     "Robbe Bierebeeck <robbe.bierebeeck@studiohyperdrive.be>"
   ],
   "scripts": {
-    "dev": "dotenv -e .env.local -- node ./scripts/generate-env-config.js && next dev -p 3200",
+    "dev-with-env": "dotenv -e .env.local -- node ./scripts/generate-env-config.js && next dev -p 3200",
+    "dev": "next dev -p 3200",
```
bertyhell, line 12: why was this changed?

reunefe: Just to avoid waiting on generating the config every rebuild like with updating the admin-core

bertyhell: but this also doesn't include the .env file
or does next 16 do that automatically for .env.local?

reunefe: The dev-with-env is something that only needs to run when the .env.local changes. After that no need to regenerate the `public/env-config.js` every time you start the dev server. I just added the new line because I ran into the issue everytime I changed something in the admin-core causing me the need to completely restart the dev server on the client. The main delay was the recreation of the env-config so to bypass that I just added a new script, that's all.

